PR #1, opened 2015-03-18
Added Swiftype search service

kovyrin:
Swiftype is a hosted search solution (has both API and crawler for ingesting content). We have a free plan with a single search engine and up to a 1000 documents allowed plus open source projects get free upgrades to the Premium level.


kovyrin:
Rebased to fix conflicts

PR #5, opened 2015-03-19
added nodequery

necenzurat:


ripienaar:
thanks

ripienaar:
hmm, maybe it's just early and I need coffee but I can't see this offer on their site?

ripienaar:
Ah, no this is 'Free during beta', I am not sure that qualifies for what I had in mind

PR #10, opened 2015-03-19
Slack

agh:
Feels like it fits under here and they have a 'Lite' plan which is free (slack.com/pricing) and perfectly suitable for getting started with limitations placed on the # of integrations and # of messaged retained in the search index.

About 60% of the teams I'm joined to are using the Lite plan and don't pay. Seems to work for them.

Thoughts or comments?


ripienaar:
perfect thanks

PR #11, opened 2015-03-19
add phplist in email category

doc75:
Add mention of phplist.com which propose hosted version with few mails.
Of course self-hosted gets unlimited access.

I let you decide if it is interesting to add it to the list.


ripienaar:
thanks :)

PR #25, opened 2015-03-30
add Scalingo in the list of PaaS

yannski:
added Scalingo in the list of PaaS. Disclaimer: I'm the CEO of Scalingo. Hope it's useful.


ripienaar:
nice, thanks

PR #26, opened 2015-03-30
Add link to comparison of CI services

ligurio:


ripienaar:
hmm, not sure it really fits the bill but will include it as it lists several free ones, but cant guarantee i wont cull it when I eventually make the website for this and thoroughly review them

thank you though :)

PR #45, opened 2015-03-30
Add Google Cloud services.

victorm95:
Correction for Google Compute Engine and add other Google cloud services.


ripienaar:
thanks

PR #54, opened 2015-03-31
Add github to the issue tracking

JonTheNiceGuy:


ripienaar:
I think if possible I want to avoid dupes like this - there are some already though so I'll merge this one.  But when I build a site I will have to figure out a way to put the same link in multiple categories cos this will soon become a nightmare.

thanks :)

JonTheNiceGuy:
How about adding tags: Github [issues, dvcs, wiki], c9.Io [ide] then filter by tags?

ripienaar:
Am doing something along those lines for the site I am developing yes

---
R.I.Pienaar

PR #55, opened 2015-03-31
Added owncloud

JonTheNiceGuy:


ripienaar:
hmm, I think aerofs isnt the same as owncloud as owncloud wants you to install a server its more in line with just downloadable free software - this means I'd end up recreating freshmeat in this list which is not the intention.

aerofs is just a bunch of clients using their closed source system to build a dropbox, so think I'm gonna decline this one sorry

PR #56, opened 2015-03-31
Vagrant box indexes

JonTheNiceGuy:


ripienaar:
not sure this fits the bill but will keep it in and decide about it later

PR #60, opened 2015-03-31
Moved Sentry to right category.

nadavshatz:


ripienaar:
thanks - not the end of the world, this is for now just raw list of links, a new site will be made with better UX and then I'll make sure its all good, but thanks for the fix :)

nadavshatz:
Got you, no problem then. Thanks for merging anyway :)

PR #62, opened 2015-03-31
Added Nitrous.IO 

ParkinT:
A great tool for Development, Testing and Collaboratio...n (Pair Programming) all browser-based.
[Here an introductory Video](http://goo.gl/J1Zbsg) I created for Learnable.com


ripienaar:
nice one

PR #65, opened 2015-03-31
Add Crashlytics to Crash section

vnavarro:


ripienaar:
not sure mobile dev falls within what I had in mind, other mobile dev related things here looked a bit more generic than this one?

vnavarro:
@ripienaar I disagree, since getsentry is a similar service and is added to the list. But is your point "generic services" exclusively implemented on server side?

ripienaar:
So the aim of the list is to collect a bunch of links that infradev people - or devops people or similar - would find useful.  

Of course, sure we could write a mobile app for something and crashlytics would be handy, I suspect though the amount of devops types who write mobile apps and would benefit from a service used only by app devs is probably slim.  

Getsentry though is not just something that's pure mobile, you can use it to track all sorts of things, things devops types would care for.

vnavarro:
Ok, I think I got the point. If I remember anything that falls into that case and is not already there I get back to this list. 
BTW nice job :wink: its indeed helpful repo.

ripienaar:
nice one thanks :) sorry for being picky but there seems to be a MUCH larger selection of free stuff than I could ever have guessed lol

PR #67, opened 2015-03-31
Put koding next to cloud 9

rwstauner:
Very similar.
Take list of features straight from their site (https://koding.com/Pricing)


ripienaar:
thanks

PR #69, opened 2015-03-31
Added Rollbar logging service

tonglil:


ripienaar:
this one is already in unfortunately, thanks for the contribution though

tonglil:
Ah, didn't see it below. No problem!

PR #85, opened 2015-04-01
Clarify Rollbar free plan

brianr:
The previous text, which I'm guessing came straight from our pricing page, confused a few people who asked us if the free plan is only allowed to be used on personal projects. Hopefully this text will be more clear.

Thanks!


ripienaar:
nice, thanks for improving that

brianr:
Np, thanks for the quick merge!

PR #93, opened 2015-04-01
Add package build systems

domcleal:
Package build systems and repository hosting sites should be particularly useful for small OSS projects.


ripienaar:
thanks

PR #96, opened 2015-04-01
Added Telerik platform to PaaS

vdininski:


ripienaar:
bit borderline if it fits the bill, might cull it when i go through all of this soon but for now its in

PR #100, opened 2015-04-01
add apiary

jbouzekri:
I don't know if I have put it in the right section.


ripienaar:
yeah seems allright, thank you.  Can you mention about its free tier like - 'ulimited users, 1 admin with limitations' or something

ripienaar:
thanks

PR #105, opened 2015-04-01
Adding Crashlytics

quinnjn:
Adding crashlytics as a free crash reporting solution for Android and iOS


ripienaar:
Thanks, I don't feel this one fits the scope of the list, see discussion here https://github.com/ripienaar/free-for-dev/pull/65

PR #107, opened 2015-04-01
Update README.md

jmelfi:
Add in Gogs, the Go Git Service for self-hosted go platform similar to gitlab


ripienaar:
mainly this is a list of hosted commercial services with a free for dev or opensource option, not a list of opensource (not rebuilding freshmeat in other words)

PR #109, opened 2015-04-01
Added Popular Text Editors

coltonfitzgerald:
If we are going to list browser based code editors, it only makes sense to list the popular desktop ones as well. What do you think?


GromNaN:
Sublime Text is not free to use; you must purchase a license even if it's usage is not restricted.
> Sublime Text may be downloaded and evaluated for free, however a license must be purchased for continued use.

ripienaar:
would prefer not to list downloadable things, mainly focussing on services else we'll soon have recreated freshmeat

PR #116, opened 2015-04-02
Adds errorception

boopathi:


TraderStf:
free only for 30 days... :-/ too bad.

ripienaar:
then i cant add it :)

boopathi:
Ah. Had a doubt whether to send this because of the 30 days thing. Cool. Let this not go then.

Issue #124, opened 2015-04-02
devport link leads to invalid SSL cert & error page


evaryont:
As the title says. The SSL cert is for the wrong domain (ghperf.com?) and if you click through, you get an error message "Not Found on Accelerator".

HTTP works, though.

PR #125, opened 2015-04-02
https doesn't work with devport (which they should fix...)

mwharrison:


ripienaar:
thank you, resolves #124

PR #129, opened 2015-04-03
Added RightScale

arangamani:
RightScale is a Cloud Management Platform and offers free trial.


ripienaar:
thanks

TraderStf:
too bad it is just a 60d trial, that would have been great!

ripienaar:
it seems - their page is crap tbf - that they always give you some amout of free time?

ripienaar:
I dont know, their page sux, so just going to take the cautious route and remove it, sorry

PR #140, opened 2015-04-04
Add Sentry

GromNaN:
Sentry: Track exceptions with modern error logging for JavaScript, Python, Ruby, Java, and Node.js

https://getsentry.com/welcome/


GromNaN:
Already there.

PR #141, opened 2015-04-04
added sendloop.com to email section

emir:
2,000 subscribers and 10,000 email delivery every month is free on https://sendloop.com/


ripienaar:
thanks

PR #147, opened 2015-04-05
Added new Freemium service Cognalys,Inc for mobile number verification

anishmenon:
I think this will help developers . www.cognalys.com


ripienaar:
thanks

PR #149, opened 2015-04-05
yubico.com two-factor authentification

TraderStf:


ripienaar:
this is just libraries to use ubikeys, you still need to buy the keys etc, dont think it fits

TraderStf:
Some of their libs are not requiring their key
https://developers.yubico.com

ripienaar:
these are just libraries, sorry its just not suitable

PR #152, opened 2015-04-05
Adds Skylight to Monitoring section

tomdale:
Hey, this is really cool! I've added Skylight, a tool for monitoring the performance of Rails apps, which is free for the first 100k requests. Thanks!


ripienaar:
thanks

PR #158, opened 2015-04-05
Add Contentful

SaschaKonietzke:
Add Contentful as a Content as a Service provider (CaaS)

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,8 @@ If you're not inclined to make PRs you can tweet me at ```@ripienaar```
 ## BaaS
   * https://www.parse.com - Mobile backends, free plan has 30 requests per second, with 20 GB of file and database storage, as well as push notifications for up to 1,000,000 unique recipients.
 
+## CaaS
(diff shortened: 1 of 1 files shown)


ripienaar:
probably more accurately described as CMS rather than ever more nebulous *aaS, but am not keen to just add more niche catagories so can you pop this one into PaaS?

thanks

SaschaKonietzke:
Sounds good. I've moved it to BaaS now, which seems to fit a lot better than PaaS (more the Heroku kind of platforms). I hope this is ok, let me know if I should change it.

ripienaar:
ah looks like its not a clean merge now, do you mind rebasing?

SaschaKonietzke:
Not the git pro unfortunately... do you mind if I create you a new pull request with only one commit? Or do you have an easy command for me to execute? Thanks and apologies for the back and forth.

ripienaar:
thats fine too, I'll hold off merging things till you do :)

PR #161, opened 2015-04-05
adding mailstache for custom email domains

burningTyger:
Free Google apps is gone, zoho might not be everybody's first choice here is mailstache.


ripienaar:
hmm, other than just a way to get email, how is this a developer tool? Other things here are stuff like list managers or subscription managers, etcetc, things dev would use to build a app.  But just getting email?  Not sure it's fitting the bill

burningTyger:
feel free to close it :-) I was thinking of the time saver you might need if you start a new shop and don't feel like building your own server.

ripienaar:
I'll add it, but might cull it when I put together the web site for this list :)

thanks!

PR #166, opened 2015-04-06
Adding logstash in the list

harryjoy:
Added Logstash (http://logstash.net/) in log management list. 
Its free and open sourced with Apache 2.0 License.


ripienaar:
list is for *aaS offerings, ie. things that are hosted not things you install on premise

PR #176, opened 2015-04-06
Added Helpmonks - free for open source projects

thenitai:
Added Helpmonks which is free for open source projects


ripienaar:
thanks

PR #179, opened 2015-04-06
Add Firebase as BaaS

metaperl:


ripienaar:
this one is there already unfortunately

PR #186, opened 2015-04-07
added firebase

reflog:


ripienaar:
it's already there, thanks though :)

PR #187, opened 2015-04-07
Added Geeklist git repositories

GabiAxel:


ripienaar:
dont think this one really fits the bill, seems more as a side feature to recruitment site

GabiAxel:
Nevertheless this is a functioning hosted GitLab instance. I have been using it for a while and it never felt like a side feature.

PR #188, opened 2015-04-07
Added Google Cloud Datastore

GabiAxel:


README.md
```diff
@@ -249,6 +249,7 @@ Table of Contents
    * https://parsehub.com/ — Extract data from dynamic sites, turn dynamic websites into APIs, 5 projects free.
    * https://import.io/ - Easily turn websites into APIs, completely free for life.
    * https://redsmin.com/ - Online real-time monitoring and administration service for Redis, 1 Redis instance free
+   * https://cloud.google.com/datastore/ - High scale NoSQL datastore, free daily quota: 1GB storage + 50K reads + 50K writes
```
ripienaar: this is already listed under PaaS as part of the larger google cloud platform entry

GabiAxel:
Yes, but since Cloud Datastore can be used as a standalone service outside of App Engine via REST API, I thought it might be worth adding it under DBaaS.

ripienaar:
if we go by that we'd have every PaaS all over the list in all categories, AWS especially and it wouldn't make things easier in the end.

When I made the website based on this raw data I'll look at a better way to show one service that relates to multiple categories

Issue #181, opened 2015-04-06
Localhost tunnels 


rscafi:
Could someone list alternatives like [ngrok](https://ngrok.com/)?

ripienaar:
If they are hosted services with a free tier sure send a pr 

---
R.I.Pienaar

PR #201, opened 2015-04-07
added ngrok.com (tunnel service)

sknebel:
This section seemed to fit best (?)
Allows to expose locally running services under a public URL by building a tunnel to their service


sknebel:
(Fixes #181)

ripienaar:
thanks

PR #202, opened 2015-04-07
Bitnami: moved & changed description

sknebel:
Bitnami only manages instances, you still have to pay your IaaS provider

see https://wiki.bitnami.com/Bitnami_Cloud_Hosting/FAQ#How_much_does_it_cost.3f


ripienaar:
thanks

PR #206, opened 2015-04-07
added: sumologic.com and appdynamics.com

pcheliniy:
Good services :)


ripienaar:
thank you

TraderStf:
May be mistaken but does not seems to have all the limitations liste in FFD readme

http://www.appdynamics.com/lite/

pcheliniy:
@TraderStf If we're talking about a Trial, You are right. But i was talking about Lite versions. At first glance, so as You describe, but here http://www.appdynamics.com/pricing/#compare more detailed information.

PR #209, opened 2015-04-08
Add shrinkray.io to Media Processing

mwarkentin:


ripienaar:
thanks, mainly looking for commercial services that have free tiers but will put this one in, might cull it later

PR #223, opened 2015-04-08
Update README.md - added bip.io

tuddman:
added bip.io to #BaaS and alphabetized the BaaS section.


ripienaar:
hmm, hard to see from their site but do they provide a hosted version of this service or is it only self host?

This list is for hosted apps so just checking

tuddman:
Its both.  Hosted or anyone is free to fork it and host it themselves.

ripienaar:
thanks

PR #224, opened 2015-04-09
Add Heroku Postgres

Joshua-Anderson:


ripienaar:
I'd really rather not list every feature of every PaaS it will make things worse, I know there is some of this going on already but I am going to cull them out and find a way when building the website to represent these features

so gonna not merge this one sorry, thanks for the effort though

PR #229, opened 2015-04-09
Added SAP Hana Cloud Platform to PAAS

jpenninkhof:


ripienaar:
is it free for ever with those limits or some shorter trial? cant see it on their site clearly

jpenninkhof:
Hello,

It is indeed a trial/developers license that won't expire. In legal lingo,
SAP calls it a free and perpetual license.

Details about what you're entitled if you sign up can be found here:
http://scn.sap.com/docs/DOC-56411

Best regards,
Jan

ripienaar:
If that's not clear right on the signup page I am not going to include this one I think. All the others on the list have clear guidance right where you would expect it. Those that don't I also did not merge - generally they say they will come back when they fixed it. 

---
R.I.Pienaar

jpenninkhof:
Not sure if I understand you correctly... It actually clearly states that it's free:
image

I've also had a look at the other PAAS services. It seems that they only mention a time limit when there is one. The ones that don't have a time limit, don't mention it. SAP's developer accounts are in the latter category.

If there is anything else you need mentioned, please let me know. Perhaps I can let them know...

ripienaar:
I guess the others include details of the limitations upfront, ie. the information in the 2nd link you provided.

Maybe just link to that send link as well in your text in the PR

jpenninkhof:
Ah, got you there. I believe they made a generic page for their regular customers as well as their trial site. That may be why they don't spell out the perks and limitations of the developer account. I like the suggestion of linking to the other page though. Let me get that done.

jpenninkhof:
As discussed in the previous pull request, the link to SAP HCP has been changed to a page that contains much more info on what a developer accounts entitles you to.

ripienaar:
thanks

jpenninkhof:
Awesome, thanks!

PR #237, opened 2015-04-11
Update README.md: change M$ to MS

souvik1997:
Using "M$" is not very professional.


ripienaar:
damn, missed this one, caught a few others, totally agree with you

PR #239, opened 2015-04-11
Added Neocities

dirkk0:


ripienaar:
hmmm, think I am going to reject this one, I dont get the feeling developers really is the target audience when you compare with the other stuff listed here like say github pages - this probably goes for some others on the list I'll cull too soon

thanks for contributing though, really appreciate it

PR #240, opened 2015-04-12
Added stripe for payment integration

VikramTiwari:
not free, but sensible pricing and development ease


ripienaar:
List is only for free sorry. 

---
R.I.Pienaar

PR #246, opened 2015-04-14
Add Firebase to BaaS

steren:


ripienaar:
already listed under PaaS

PR #247, opened 2015-04-14
Adding freehostia.com

seoulgithub:
Very good hosting service for php as well as static html.


ripienaar:
this is a truly horrific website, just not going to add it sorry.  There are millions of free php webhosts all catering to prosumers who want to run some php thing they got on the net but the ones listed here are much more modern *aaS offerings.

andrewbrereton:
:+1:

PR #248, opened 2015-04-14
Added sslcheck.globalsign.com

Eun:
Basicly another SSL configuration analysis service.


ripienaar:
thanks

ApfelUser:
It is just a skin for the Qualys SSL Analyzer, you can see that if you scroll to the bottom and look at the footer.

ripienaar:
Ah did think it was a pretty good clone :) thanks for pointing that out I will revert it a bit later today 

---
R.I.Pienaar

PR #251, opened 2015-04-14
Added Shelly Cloud on the PaaS list.

smt116:


README.md
```diff
@@ -225,6 +225,7 @@ Table of Contents
   * http://azure.microsoft.com/ - MS Azure gives $200 worth of free usage for a trial
   * http://hpcloud.com/ - $300 credit over 90 days.
   * https://appharbor.com/ - A .Net PaaS that provides 1 free worker
+  * https://shellycloud.com/ - Platform for hosting Ruby and Ruby on Rails apps. Shelly Cloud gives €20 free credit
```
ripienaar: hmm, that's not even 30 days of one tiny instance, dont think that fits the bill for this list.  There are other free tiers listed here but they tend to be a bit more substantial.

smt116: €20 is enough for 30 days of running small virtual server. This is the same time period as for bluemix.net.

ripienaar: I'll add it, but be aware when I make the website to go with this data i'll likely cull such short trials out

ripienaar:
thanks

README.md
smt116: Ok

PR #254, opened 2015-04-15
Items added to list

PierceMcGeough:
Beanstalkapp to Source Code Repos
Sublime Text to IDE and Code Editors

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Table of Contents
   * https://hub.jazz.net/ - Unlimited public repos, private repos free for up to 3 accounts.
   * https://visualstudio.com - Free unlimited private repos (Git and TFS) for up to 5 users per team
   * https://assembla.com - Free repo hosting in a free plan.
+  * http://beanstalkapp.com/ - Multi user Git and SVN hosting (Free 1 user 1 repository account)
 
 ## Tools for teams & Collaboration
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -356,6 +357,7 @@ Table of Contents
    * https://www.visualstudio.com/en-us/products/visual-studio-community-vs.aspx - Visual Studio. Not only for Windows and .NET
    * https://codio.com/ - Codio is a cloud-based computer programming platform for universities, schools, and developer professionals.
    * http://www.stackhive.com/ - Cloud based IDE in browser that supports HTML5/CSS3/jQuery/Bootstrap
+   * http://www.sublimetext.com/3 - Sublime Text is a sophisticated text editor for code, markup and prose.
```
ripienaar: i believe the other here are hosted or very tightly coupled with a hosted component, sublime is mainly a download standalone app am I right?

If that's the case I'd rather not list it

PierceMcGeough:
Yes it is downloadable but has many plugins that can be installed to make life easier. Jetbrains PHPStorm is downloadable and standalone and it is on the list. We all use both at work PHPStorm for backend and Sublime for the front.

ripienaar:
I'll rather remove phpstorm then, this list is for *aaS only as per the top of the readme and description

PierceMcGeough:
Fair enough then.

One other thing
https://www.atlassian.com/opensource/overview  should really be
https://www.atlassian.com/software/jira

I think first link is a bit misleading. Need to fish around to get to the Jira page from there

ripienaar:
at the time thought the current link was easier to find the open source offer but yeah just linking to https://www.atlassian.com/ would be best since they give all their software away for free for OSS

PR #255, opened 2015-04-15
Adding ThoughtWorks Tablero Project to Issue Tracker section

flmmartins:


ripienaar:
thanks for your contribution! unfortunately the list is for *aaS software, ie. ones that are hosted and not self install so unfortunately I won't be merging this one

thanks though

PR #264, opened 2015-04-18
liferay.com

TraderStf:
http://www.liferay.com/downloads/liferay-portal/overview


ripienaar:
list is not for self hosted software, sorry

PR #265, opened 2015-04-18
https://ellislab.com/expressionengine

TraderStf:
https://store.ellislab.com


ripienaar:
as other PR, list is not for self hosted software

PR #275, opened 2015-04-22
bitrise.io added to CI / CD section

viktorbenei:


README.md
```diff
@@ -94,6 +94,7 @@ Table of Contents
 
 ## CI / CD
 
+  * https://www.bitrise.io - CI & CD service for iOS projects. Free for typical personal use.
```
ripienaar: Thanks for the contribution.  The list is aimed at devops and infradev people so I've been avoiding pure mobile development tools - there are some on the list but I thought they'd have broader appeal too, this one looks pretty niche so don't think it will fit the bill tbh

Thanks for the effort though

viktorbenei: No problem at all, thank you for the explanation!

PR #279, opened 2015-04-23
Nitrous.io change of policy.

shakna-israel:
Nitrous.io are phasing out the free tier as it stands.

Instead, the new free tier will be 2 hours per day.


ripienaar:
thanks for the update

PR #280, opened 2015-04-23
ptengine.com web analytics

TraderStf:


README.md
```diff
@@ -372,6 +372,7 @@ Table of Contents
  * https://usabilityhub.com - Test designs and mockups on real people, track visitors. Free for one user, unlimited tests.
  * https://gosquared.com - Track up to 1,000 data points for free.
  * https://mixpanel.com - Free 25000 points or 200000 with their badge on your site.
+ * http://ptengine.com - Free 20000 page views/month, one domain and one heath map.
```
ripienaar: does this site work reliably for you? on 2 computers with 3 browsers it renders blank pages at random times, sometimes page loads then it clears again to blank page.  think they have some bad JS

TraderStf: I do not try all sites I put on FFD :-)))

they exist since mid 2014, so should work...

perhaps because they are in japan

got this by whois

ripienaar: ah, its also very slow, dont think I'll add it

Issue #283, opened 2015-04-24
is realm.io really DBaaS?


themasch:
For me it looks like a local DB/ORM Framework, not like DBaaS. 
https://realm.io/

ripienaar:
yeah you're right, does not belong on the list, must have slipped through my reviews.  list should be for *aaS only.

Want to do a PR to kill it?

PR #284, opened 2015-04-24
Update README.md

themasch:
removed realm.io because its not a service but a framework. (#283)


ripienaar:
thanks for sorting that out

PR #285, opened 2015-04-24
Added Binary repositories section for bintray

adriens:
Added bintray

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,7 +346,7 @@ Table of Contents
    
 ## Binary repositories
 
-   * https://bintray.com/ - binary repository system (maven, .deb and .rpm). Repo can be registered for each system, making updates easier.
+   * https://bintray.com/ - binary repository system (maven, .deb and .rpm ... and even docker). Repo can be registered for each system, making updates easier.
    
 
 ## IDE and Code Editing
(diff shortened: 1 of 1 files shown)


shakna-israel:
* You didn't add ```Binary Repositories``` to the TOC
* Isn't this more ```Package Distribution```?

ripienaar:
It does probably need a category yes but I am not keen on adding single entry cats - going to merge the ones there are now

So maybe add 1 or 2 of the other free for oss ones at the same time?

adriens:
"It does probably need a category" > Yes, that's exactly what i thought
"but I am not keen on adding single entry cats" > totally agree with you, but i didn't really know where to put it, there is a Package Distribution that seems appropriated while bintray does not compile package, just store them, but in my opinion, if it had to be stored somewhere it would rather be Package distribution. Yes, sorry, i did not add it to the TOC

adriens:
New PR created to put bintray in Package Build Systems  section.

ripienaar:
hmm, might be losing it but I am not seeing another PR? can you link to it?

adriens:
Yep, sorry : here is the one i'm talking about : 

https://github.com/adriens/free-for-dev/commit/87ed6a0fe28fe025f620db894983b73a46b3090e

ripienaar:
as commented there, you need to make a PR if you want me to merge it.  The text is also a bit on the long side.  Going to close this one waiting for the new PR

PR #287, opened 2015-04-25
Update README.md

githubashto:
Added email services.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,11 @@ Table of Contents
   * https://postmarkapp.com - First 25,000 emails are free
   * https://www.zoho.com/mail/ - Free Email management and collaboration for up to 10 users.
   * http://moosend.com/ — Mailing list management service. Free account for 6 months for startups.
+  * https://sendsay.ru/ — 200 subscribers and 1,000 emails per month for free
+  * http://smartresponder.ru/l_en/ — 1,000 subscribers and 50,000 emails per month for free
+  * http://pechkin-mail.ru/ — 100 subscribers and 500 emails per month for free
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -209,6 +209,11 @@ Table of Contents
   * https://postmarkapp.com - First 25,000 emails are free
   * https://www.zoho.com/mail/ - Free Email management and collaboration for up to 10 users.
   * http://moosend.com/ — Mailing list management service. Free account for 6 months for startups.
+  * https://sendsay.ru/ — 200 subscribers and 1,000 emails per month for free
```
ripienaar: might be missing it but cant find a english page for this, its probably useful but way too niche to be a russian language only one

ripienaar:
the 2 english language ones will fit the bill, but I think the pure russian language sites are too niche, if you can tweak the PR to remove those - or maybe there are english links I cant find - that'd be great

TraderStf:
If you allow me, I added few sites with only non-english language as this one, some with few different languages and or locations, why reject them?

GitHub has a lot of projects which are in a single non-english language and/or concerning a single country.

For me it is the same :-) I really don't care.
Have a nice sunday.
Stf

ripienaar:
You did? I recall blocking some pure non English ones but don't recall adding ones with no English part

As per my comments. Happy to add the ones that have multiple languages of course. 

---
R.I.Pienaar

TraderStf:
What I try to say is these niche ones might be useful, even if that specific.
A russian will prefer to use yandex cdn rather than one in US for its clients.
Bye :-)

ripienaar:
no response, feel free to re-open later

PR #294, opened 2015-04-28
Add Imagefly to "Storage and Media Processing"

joeshaw:
Imagefly is a responsive image service that does image resizing, transcoding to formats like WebP and JPEG XR, and other image optimizations.  It's free for 100 MB/month.


ripienaar:
thanks!

PR #298, opened 2015-05-01
added baas.io in BaaS section

imyaman:


ripienaar:
hmm, as it's only in korean I cant evaluate if its free or not or in beta or whats going on.  It also severly limits the general usefulness - and this list should serve a wide audience.

So I've not been taking sites that are only available in a non english language and while I cant confirm this is the case with this one I also dont take beta sites that dont show future paid versions will have a free tier for beta users

thanks for the effort but at least till there's an enlish site I cant evaluate it and accept it

imyaman:
Can I add README.ko.md?

ripienaar:
no sorry, the readme thing is only temporary this data is being moved into a website as i get a chance so that wont help :(

PR #299, opened 2015-05-01
Include Google App Engine under free email providers

johnmellor:
References:
- https://cloud.google.com/appengine/docs/quotas#Mail
- https://cloud.google.com/appengine/docs/quotas#Requests
- https://cloud.google.com/appengine/pricing#cost_resource


ripienaar:
thanks, but I prefer to list these big cloud companies with 10s of services just once else it will never end.

am working on a site that will support tags or something similar that should make it better to do this, but while this data is in this giant readme format it will just make it worse :(

PR #300, opened 2015-05-01
Added Stripe to Payment / Billing Integration

alotaiba:


ripienaar:
does it have a free tier?

alotaiba:
The only information I see is "2.9% + 30¢ per successful charge". I don't think they offer anything for free besides "No monthly fees, No refund costs, No hidden fees".

ripienaar:
then it doesn't seem suitable for a list of software with free tiers :(

sorry

PR #305, opened 2015-05-01
Add Visual Studio Code to IDE and Code Editing section

aaronkor:


ripienaar:
Thanks for the PR. The list is for hosted software as a service and not for download only software so not sure I will take this one sorry

---
R.I.Pienaar

aaronkor:
Oh missed that part. Sorry. Closing.

PR #310, opened 2015-05-03
added data mining cloud services

amaboura:
These are few Data mining cloud services i wanted to add.

Thanks for the list!

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,3 +410,6 @@ Table of Contents
 
 ## Data mining
   * http://www.monkeylearn.com/ - Text mining in the cloud, 1,000 queries for free per month.
+  * https://wit.ai/ - NLP for developers
+  * http://azure.microsoft.com/en-us/services/machine-learning/ - Predictive Analytics in the cloud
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -410,3 +410,6 @@ Table of Contents
 
 ## Data mining
   * http://www.monkeylearn.com/ - Text mining in the cloud, 1,000 queries for free per month.
+  * https://wit.ai/ - NLP for developers
+  * http://azure.microsoft.com/en-us/services/machine-learning/ - Predictive Analytics in the cloud
```
ripienaar: nice, but I dont want to list every service that these huge clouds have - else amazon ML belongs here too etcetc, it never ends.

Am working on a web site for this data which might have tags till then I think it's best if we dont duplicate things in this format.

rest looks good, do you mind removing this one?

ripienaar:
thanks, sorry for the late merge - github doesnt notify when you update a branch without commenting so I missed that you fixed it

PR #313, opened 2015-05-04
Add Docker

mychaelgo:


ripienaar:
The list is for commercial *aaS that has free tiers for debs. The 2 items does not fit the scope sorry 

---
R.I.Pienaar

mychaelgo:
@ripienaar ok :)

PR #316, opened 2015-05-04
Updated SendGrid free availability

aaron7pm:
Microsoft Azure offers the same 25,000 emails per month for free with SendGrid. All you have to do is make an instance from the Azure Marketplace.


ripienaar:
thanks

PR #319, opened 2015-05-05
GitHub gives education packs to high schoolers too

haneefmubarak:
I currently attend high school as an 11th grader. I sent in verification to GH and they granted me an education pack.


ripienaar:
thanks

PR #326, opened 2015-05-08
Remove http://www.hall.com

SHoar:
Hall has been acquired by HipChat. No new user accounts are available at this time.


ripienaar:
thank you

PR #327, opened 2015-05-09
added trailing slashes to unify urls

nadsumatal:
I added trailing slashes to urls to improve consistency. Reasons for adding slashes and not removing:

 - most did have trailing slashes
 - the slash speeds up loading of the site


ripienaar:
thank you

PR #329, opened 2015-05-10
visualping.io - alert of visual changes

TraderStf:
Unusual ping tool, alert of visual changes.
in other words, it is just like you capture an area on a page and when that area visual changes, you are notified, very handy!


ripienaar:
Handy yeah but not sure how it's a developers saas tool :(

PR #330, opened 2015-05-11
add bosonnlp: a NLP api service for Chinese

axiaoxin:
NLP for Chinese

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,3 +418,4 @@ Table of Contents
 
 ## Data mining
   * http://www.monkeylearn.com/ - Text mining in the cloud, 1,000 queries for free per month.
+  * http://www.bosonnlp.com/ - NLP for Chinese.


README.md
```diff
@@ -418,3 +418,4 @@ Table of Contents
 
 ## Data mining
   * http://www.monkeylearn.com/ - Text mining in the cloud, 1,000 queries for free per month.
+  * http://www.bosonnlp.com/ - NLP for Chinese.
```
ripienaar: Hello,

I can unfortunately not take sites that does not have an english version of their site as I need to be able to evaluate the sites for suitability and maintain the list on the long run.  I'd also prefer something thats a bit wider audience than something only for one country

Besides that, no idea what "NLP" is and cant verify if this site gives anything for free so not able to add it sorry

ripienaar: ah, neuro linguistic programming I guess

axiaoxin:
Natural language processing :)

PR #332, opened 2015-05-11
added the free SSL offering from Cloudflare in ## Security and PKI

stefancocora:
As the title says I've added another free SSL offering.


ripienaar:
So cloudflare is already in "CDN and Protection", maybe just update their description mentioning the free ssl offer else it gets a bit (more) messy

stefancocora:
Ok I'll do that , closing this PR , I'll resubmit another PR.

PR #335, opened 2015-05-11
added codepen and jsfiddle, fixed bullet levels

maxehnert:


ripienaar:
these are useful but I am not sure they really fall in the SaaS category they are more little helper tools of which there are many many many of them.  I think the value here is in listing commercial services that has free tiers - there are a few here that are plain free ones here but trying to keep them to a minimum

maxehnert:
That's fair. 
The codepen one is a tiered service which has paid features not available to free users like screen sharing for pair programming. 
If you want, I can remove the links and just submit the bullet fix, or none of the above ; )

ripienaar:
ah lets leave codepen and take out jsfiddle then we're all good

maxehnert:
I think it's fixed. I haven't made many pull requests so not sure if I did that right.

ripienaar:
sweet, thanks

PR #337, opened 2015-05-12
Updated Codenvy.com entry in Readme

bmicklea-ce:
Also moved Codenvy up in the list to reflect alphabetical ordering.


ripienaar:
thanks

PR #338, opened 2015-05-13
Add cronasaservice.com

drei01:
This list doesn't contain any other online hosted cron services and cronasaservice.com provides an unlimited free trial.


ripienaar:
neat, thought we had some already.  Might make more sense under IaaS section?

drei01:
Good idea, I've moved it

ripienaar:
thanks

PR #346, opened 2015-05-14
Add evernote to the list

bogdal:
This PR adds great tool for organizing information.


ripienaar:
hmm, i can see how its useful for a developer but its not really a development tool so it doesnt feel like a good fit to me, what do you think?

bogdal:
Yep, this is not a typical tool for developers, but it helps you to collect all your project requirements, discussions and ideas in one place and share them with other co-workers. On the other hand, this is only one of its purposes. If you feel that it doesn't fit to the list, it's ok for me.

ripienaar:
thanks

PR #352, opened 2015-05-20
about quay.io (Docker) hosting

piccaso:
From what i can tell its not possible to run *containers* on quay.io, at least not with the free account I have...
Could this be a misunderstanding of some sort? (containers/repositories)

Btw.: I would love to be wrong on this one, so if there is a way to run containers there please let me know! (And ditch the PR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,9 +414,9 @@ Table of Contents
  * https://education.github.com/pack - As long as you're a student at a recognized university or US high school
 
 ## Docker Related
-### Alternate container hosting
+### Alternate hosting
 
-  * https://quay.io/ - Unlimited free public containers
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -414,9 +414,9 @@ Table of Contents
  * https://education.github.com/pack - As long as you're a student at a recognized university or US high school
 
 ## Docker Related
-### Alternate container hosting
+### Alternate hosting
 
-  * https://quay.io/ - Unlimited free public containers
+  * https://quay.io/ - Unlimited free public repositories
```
ripienaar: you cant run them, but the images are stored there yes so indeed its a bit confusing

PR #364, opened 2015-05-24
Analytics: add Amplitude.

rberidon:


ripienaar:
thanks for this and the others

PR #374, opened 2015-06-02
added coral cdn

JonW27:


ripienaar:
hmm, this is nice but I think not exactly relevant, when making the list I wanted to get commercial quality services that cater for devs either in their own exploratory hacking or perhaps for hosting demos etc - in effect making it free to gain experiences in the SaaS world without in many cases even to commit a credit card

later customer could pay for the full commercial service etc.  of course a whole heap of purely free things made the list - tend to be things that supports development workflow etc like kanban boards etc.

but when it comes to CDNs, I think what I'd consider most appropriate are commercial ones with a free tier rather than community efforts like coral is

does that make sense?

JonW27:
ok

PR #378, opened 2015-06-05
Update README.md

fartusse:
Just added Restlet APISpark platform which is an API platform to build APIs from scratch in few minutes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,7 @@ Table of Contents
   * https://www.incapsula.com/ - free CDN and DDoS protection
 
 ## PaaS
-
+  * http://restlet.com/products/apispark/ - APISpark enables any API, application or data owner to become an API provider in minutes via an intuitive browser interface.
   * https://cloud.google.com/appengine/ - Google App Engine gives 28 instance hours free, 1Gb NoSQL Database and more.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -232,7 +232,7 @@ Table of Contents
   * https://www.incapsula.com/ - free CDN and DDoS protection
 
 ## PaaS
-
+  * http://restlet.com/products/apispark/ - APISpark enables any API, application or data owner to become an API provider in minutes via an intuitive browser interface.
```
ripienaar: looks like this should go with other similar things under BaaS

fartusse: @ripienaar Makes sense, but Restlet is not only doing backend, that's why PaaS will be more global for the activity.

ripienaar: It's closer to IaaS than PaaS, regardless their overreaching marketing collateral.  You just have to look at the feature lists of others in the category to know this one is not a PaaS.

jlouvel: Hello @ripienaar I double check the other entries in the PaaS section and APISpark really seems to fit there AFAIU. Gartner recently recognized us in the specialized API PaaS category: http://restlet.com/blog/2015/05/11/restlet-named-a-cool-vendor-in-specialized-paas-by-gartner/

ripienaar:
thanks

PR #381, opened 2015-06-10
Added tunnelbroker.net & embed.ly

mayurkr:


ripienaar:
I think embed.ly is a good fit but tunnel broker isn't really what I am after - this is more about commercial offerings that has a free tier 

Could you remove it and leave embed?

Thanks!

---
R.I.Pienaar

mayurkr:
:+1:  did it, you can merge it now :)

ripienaar:
sweet, thanks

PR #389, opened 2015-06-27
Added Logstash to Log Mangement section

thomasbabuj:


ripienaar:
hey

thanks for this, unfortunately the list isn't for self hosted open source, its primary a list for hosted software as a service offerings that has a free tier for development activities.  So while logstash is great this isnt't for this list

thanks for the effort though

thomasbabuj:
got it. thanks for the update.

Issue #111, opened 2015-04-01
WebSite Tools for Free-For-Dev Directory


TraderStf:
Hello,

As I make a list/directory of sites, on totally different subjects, I have searched for appropriate tools to easily handle all these links.

Here is what remains on my list until today. I have no link with any of them.

I hope this could help you to do FFD Directory.

Stéphane.

<h3>— Arfoo —</h3>

[Arfooo](http://script.arfooo.net/) is a free and Open Source directory script.

Apache (mod_rewrite)
PHP 5.1
MySQL 4.1
Javascript, jQuery, SPL, Curl, GD, XML
MVC architecture
UTF8 support

[Arfooo 2.0 Features](http://script.arfooo.net/features/)

<h3>— eSyndiCat —</h3>

[eSyndiCat](http://script.arfooo.net/) is directory script available in two versions:
- version 1.7 is free which is maintained and kept up-to-date though with less features.
- version 2.3 is a 75 $USD one time payment for one domain with no hidden fees.

Apache 1.3 (mod_rewrite) or Microsoft IIS (ISAPIRewirite 3)
PHP 5.2
MySQL 4.1
GD, XML, Zlib, Freetype
<i>ionCube loaders will be required for version 3</i>

[eSyndiCat Features and Comparison](http://www.esyndicat.com/product/comparison.html)

<h3>— SMF Forum —</h3>

As I was thinking to list sites but also to let visitors add comments, recommendations, votes... I have also searched for forum / webboard / bulletin board tools.

I have chosen [Simple Machines Forum](http://www.simplemachines.org/) which is neat, powerful not bloated, and very secure.

SMF is free and Open Source.

Apache 2.0 (mod_rewrite) or Microsoft IIS
PHP 5.5
MySQL 4.0.18 / PostgreSQL 8.0 / SQLite 2.x (not 3.)

[SMF Features](http://www.simplemachines.org/about/smf/features.php)

ripienaar:
house keeping ancient issues

PR #397, opened 2015-07-13
Moved Opbeat to Monitoring list.

makwarth:
Since the monitoring list is getting quite long, it might be helpful to split it into multiple lists, like "server monitoring", "application monitoring", "alerting" .. Happy to help :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,6 @@ Table of Contents
   
 ## Management Systems
 
-  * https://opbeat.com/ - Release, deploy, monitor.  Free for 3 users
   * https://bitnami.com/ - Deploy prepared apps on IaaS. Management of 1 AWS micro instance free
 
 ## Log Management
@@ -162,6 +161,7 @@ Table of Contents
 
 ## Monitoring
 
+  * https://opbeat.com/ - Application performance, errors & releases. Free with 24 hour data retention.
   * https://checkmy.ws/en/solutions/free-forever-for-foss/ - Free 15 days full demo and 3 websites forever free for Open Source projects.
   * http://www.appneta.com/ - Free with 1 hour data retention
   * https://www.thousandeyes.com/ - Network & user experience monitoring. 3 locations, plus 20 data feeds of major web services free.


ripienaar:
yeah its a pain, I am hoping to get some time to develop a website where entries can have multiple tags or catagories which will help.

Will move this one for now, but yeah def need something better and I think no matter what we do this one-giant-readme just wont get any better :P

makwarth:
I actually like the simplicity of the giant readme - and the fact that it's on github. Let me know if you want help tidying up the lists :)

cheiff:
Hi,

maybe this project can help:
https://github.com/mwozniak/bookmark-manager

Regards

ripienaar:
Indeed the fact that it's just some file is I believe one of the main keys to success - its easy for anyone to quickly edit/add an entry even in the github editor

So the site I've been hacking on now and then has a very similar - but slightly more structured - flat file as  "database" to hopefully keep the easy workflow for contributors while giving higher order structure and functionality.

@makwarth help would be good, if you have a proposal it would be nice to see it outlined - just write a comment detailing it rather than do a ton of work on editing the doc and making a PR. 

Once we have an idea of where we're headed we can do work?

@cheiff thanks - but as above, I want to avoid DBs

cheiff:
Yes, I understand. I also like the flatfile solution.

If you want to work on a flat file solution with tags, etc, I'll be 
happy to help.

paulbrab:
What about storing it in json?

ripienaar:
@paulbrab JSON isnt very human friendly, if it has to come to a serialized format it'll probably have to be YAML or TOML I suppose

Waterloo:
Can't we scrap the list

paulbrab:
apologies @ripienaar I meant for use in a website instead of a db, but yes, not very friendly for editing

makwarth:
@ripienaar Been giving this some thought and I honestly think the list is working as it is. Over time, large lists should be split into sub-lists, but for now I would just make sure all entries have a concise description of the product offering. [name] - [what it does in] - [free plan conditions]

ripienaar:
@makwarth yeah, I do intend to keep it almost that simple with just a few extra fields, so you end up with a human readable README that's effectively a serialised data structure to drive a web view of same.

Thus keeping the easy to edit and search with ^F etc nature and adding the ability to present the categories better and add entries into multiple categories - this has been a very common thing in PRs, the same service in multiple categories and it will make this list massive and impossible to manage if we did it via duplication

makwarth:
@ripienaar cool :+1:

TraderStf:
I have suggest this tools long time ago.
https://github.com/ripienaar/free-for-dev/issues/111

ripienaar:
Thanks @TraderStf but as you can see from the preceding discussion ones with Databases will not work

PR #400, opened 2015-07-15
Added Design&UI and Bitrix24

mthjn:
Indeed. Forked anew, looks better now. Sorry for all the hassle btw.


ripienaar:
nice, thanks for the effort really appreciate it

PR #402, opened 2015-07-16
Update Readme.md

ankit-saxena:
Added comment for Cine.io. The service is shutting down on August 31st.


ripienaar:
Thanks - can't we just delete it? Seems little value in keeping it around

ankit-saxena:
Sent another pull request. closing this one.

PR #405, opened 2015-07-19
Added nullege.com.

levalex:


README.md
```diff
@@ -98,6 +98,7 @@ Table of Contents
   * https://www.pullreview.com/ - Automated Code Review for Ruby in GitHub, Bitbucket and Gitlab - free for Open Source
   * http://gocover.io/ - Code coverage for any [Go](http://golang.org/) package
   * http://inch-ci.org/ - Documentation badges for Ruby, JS & Elixir
+  * http://nullege.com/ - Search engine for Python source code
```
ripienaar: not sure this fits the bill, the list is more for *aaS services that form part of the infrastructure of the resulting program.  Sure a search engine like this is useful but for the list you'd rather include a *aaS that lets you build search engines as an example

thanks for the effort though really appreciate it

PR #406, opened 2015-07-19
Added Scalingo and VisualOps

Reubend:


Reubend:
Actually I'm not sure that VisualOps is a PaaS. It definitely has a free plan but I may have mis-categorized it...

PR #414, opened 2015-08-04
devopsboard.com has been added into management systems

evgeny-savitsky:
Here is another usefull tool we use and suggest others to use it in their daily work

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ Table of Contents
 
   * https://bitnami.com/ - Deploy prepared apps on IaaS. Management of 1 AWS micro instance free
   * http://www.visualops.io/ - 3,600 instance hours per month free
+  * http://devopsboard.com/ - Manage software and hardware issues aggregated from monitoring tools on single Kanban board
 
 ## Log Management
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,6 +146,7 @@ Table of Contents
 
   * https://bitnami.com/ - Deploy prepared apps on IaaS. Management of 1 AWS micro instance free
   * http://www.visualops.io/ - 3,600 instance hours per month free
+  * http://devopsboard.com/ - Manage software and hardware issues aggregated from monitoring tools on single Kanban board
```
ripienaar: does this have a free tier or is it just a trial? looks like just a trial to me which wouldnt fit the goal of this list

evgeny-savitsky: Can't say more clear, we use it free

evgeny-savitsky: Actually targetprocess, jira and so on are not free at all too

ripienaar: All I see on their pricing page regarding it is:

```Submit the credentials on the left and get your instance ready in couple of seconds. We give you 14-days trial period to use DevOps board absolutely free.```

which suggests its a trial only

evgeny-savitsky: there is hosted option, maybe they're providing it free for open-source or academic needs?

ripienaar: does not seem that way, you added it here - based on where did you see its free? if I cant see its free I cant add it

evgeny-savitsky: Ok, sorry for that

PR #416, opened 2015-08-07
add gitcafe.com

akar1nchan:
add gitcafe.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Table of Contents
   * https://bitbucket.org/ - Free unlimited public and private repos (git and mecurial) for up to 5 users
   * http://chiselapp.com/ - Unlimited public and private Fossil repositories
   * https://github.com/ - Free for an unlimited number of public repositories
-  * https://gitcafe.com/ - Free for an unlimited number of public repositories
+  * https://gitcafe.com/ - Free public repositories for developers
   * https://about.gitlab.com/ - Unlimited public and private git repos with unlimited collaborators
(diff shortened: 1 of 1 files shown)


ripienaar:
I cant seem to find details about their free tier, i see 'try for free' but does that mean it stays free or does it mean its a trial?

akar1nchan:
Free for open source and paid for private use.

ripienaar:
I need to be able to see this mentioned clearly on the site before I can add it, do you have a specific link?

yiliangt5:
@ripienaar @akar1nchan 
Hi both,
I am able to find the pricing terms of GitCafe.com but it seems that this info is in Chinese only: https://help.gitcafe.com/manuals/charge-strategy/content

Following is translated by Google:
All pay fees to use GitCoin
Each private project cost is 4 GitCoin / day
Each time you add a private project collaboration personnel costs 4 GitCoin / day
**Get a free 512 MB per account hosting space**, after the excess of the free use of space, the cost of the excess is 4 GitCoin / 512 MB
30 GitCoin values 1CNY

That's to say each user has 512 MB quota for hosting their public repos for free.

Hope that will help.

ripienaar:
OK thanks.  I am afraid then I can't add this one.  It's not that I don't believe you but I have to be able to easily verify what it says, in the past several chinese and russian sites were submitted and I then too only accepted ones with english pages.

There seems to be a lot of chinese resources that would fit the bill and no doubt a separate list could be made but for this one i'll have to stick with things I can verify and broadly usable which unfortunately means being in English

PR #420, opened 2015-08-12
Suggesting addition of Giant Swarm.

kordless:
Thank you for building this list!


ripienaar:
Looks nice the problem is the 'free while in alpha', question is then what? I can only add things that clearly commit to retain a free tiers for developers

ripienaar:
they clarified their future pricing, won't have a free tier so i am closing this one

PR #421, opened 2015-08-13
observium.org monitoring

TraderStf:


ripienaar:
this appears to be self hosted software and not a *aaS offering

TraderStf:
Then what 😜

As there is free hosting for dev, this will become an *aas. it is not a desktop computer program.

It is a combo 😇

Agree? 😘

ripienaar:
this list is for free tier *aaS.

PR #424, opened 2015-08-18
Added https://filehero.io/

mpociot:


ripienaar:
thank you

buchmiller:
This website appears to be dead now. It should be removed from the list.

ripienaar:
Thanks, feel like sending a PR to kill it?

PR #426, opened 2015-08-19
added getstream.io - a newsfeed and activity stream PaaS

DanielBodnar:


ripienaar:
nice, thanks

DanielBodnar:
I'm still a little unclear the difference between PaaS and BaaS - do you think this would be better filed under BaaS?

ripienaar:
yeah actually I think so, naming things is hard :P

ripienaar:
PaaS is like Heroku, BaaS is like things you might find listed as plugins for Heroku here https://elements.heroku.com/

roughly, but really it's a wavy line

DanielBodnar:
I'll make that change and submit another pull request

PR #428, opened 2015-08-19
Added section.io - a globally distributed varnish cache solution

rghv:
Added www.section.io -  a globally distributed complete Varnish Cache solution


ripienaar:
sweet, thanks

PR #429, opened 2015-08-21
Firebase pricing update

ericthornton:
Firebase has updated their free tier pricing, doubling the max
concurrent users, database bandwidth and database storage.

https://www.firebase.com/blog/2015-08-19-announcing-lower-simpler-pricing.html


ripienaar:
nice, thanks

PR #436, opened 2015-08-29
Add Coverity Scan

pcolby:
Coverity Scan is the free-for-Open Source SaaS version of Coverity's static analysis product.


ripienaar:
thank you

PR #450, opened 2015-10-03
Adding appenlight

daneoshiga:
I've been using appenlight for many months now in the free plan setting, but they removed all information about pricing and plans from site, what remains is [this discussion](http://discuss.bootstrapped.fm/t/hi-im-marcin-from-appenlight-com/1711/2) with the following text:

Exceptional services start free
So you get one month of Pro plan on signup (No CC required).
After 30 days you can still use our free forever plan!

I believe they just removed the Pro plan for now, so all accounts are free.


ripienaar:
hmm, that's weird, hard to see how a thing like this can be totally free for very long, how does it survive?

It's great and all that its free but I wonder if its worth adding something that does not seem clear about its intentions - so far I avoided adding them.

the intention of this list is a consultant dev can use these services - or easily explore their features - and then should they decide to use them their customers can opt to pay for a proper account.  So a service that's free for dev but with a commercial option is really well suited which is what most of the ones here are.  But this one just make me wonder if it will be around much or what their plans are?

daneoshiga:
Understood, since I use it I can check if they become more clear about their plans, will reopen this when that happens, thanks :)

PR #454, opened 2015-10-08
airtable.com

TraderStf:


TraderStf:
It seems a nice tool to manage FFD

ripienaar:
nice, weird but nice :)

PR #460, opened 2015-10-10
newrelic.com 4 types of monitoring

TraderStf:


ripienaar:
Think it's better to just list them once

PR #463, opened 2015-10-12
Added blackfire

thomasbabuj:
Web App Profiler


ripienaar:
hmm, think we should probably just add this to the monitoring section, it's tangentially related - there does not seem many of these out there for free so I am not keen on a single entry catagory

thomasbabuj:
Got it and move the link under monitoring. Thanks.

ripienaar:
thanks a lot :)

PR #483, opened 2015-10-28
sandstorm.io

garrmark:
Sandstorm.io is an open source hosting platform for personal/private instances of web apps. Users can upload and install arbitrary software through a simple app-store-like web interface. In addition to better privacy, this is the only way to make Open Source web apps viable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,7 @@ Table of Contents
   * https://viaduct.io/ - 350 MB of RAM and a 1 GB database for free
   * http://pagodabox.io/ - Small worker, web server, cache, and database for free
   * https://www.cloudandheat.com/en/paas.html - 128 MB of RAM for free, includes support for custom domains for free
+  * https://apps.sandstorm.io/ - An open source platform for servers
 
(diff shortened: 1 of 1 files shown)


ripienaar:
hard to make out exactly what this is - looks like they have a self hosted option and some option for managed hosting.  But there are no pricing for the managed hosting version to be found.

I've in the past not accepted self hosted software so if the free part is self hosting I dont think it would fit here.

Can you point me at pricing info for the managed hosting part?

shakna-israel:
It's accessible once you sign up... But that's annoying as hell.

The prices listed in the [release announcement](https://blog.sandstorm.io/news/2015-08-31-oasis-beta-launch.html) are still the same as far as I'm aware.

ripienaar:
OK, then as nice as this looks I dont think it really fits the bill for this :( there's no ongoing free tier (during beta only...) and the only free option is self hosted

sorry :(

garrmark:
No worries, I couldn't find pricing either so I thought there was a free option.

PR #484, opened 2015-10-30
sumome.com

TraderStf:


ripienaar:
why remove all the trailing ```.```? the one you're adding also have a trailing ```.``` but you're removing others :)

TraderStf:
Doooeee, i'll remove mine another day

PR #486, opened 2015-11-03
Added Wistia under tools for teams section

thomasbabuj:
Video hosting with viewer analytics, HD video delivery


ripienaar:
thanks

PR #489, opened 2015-11-03
add asana for Team Collaboration

aruntakkar:
your list is great we can add asana free team collaboration Also what's your take on this

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Table of Contents
 
 ## Tools for Teams and Collaboration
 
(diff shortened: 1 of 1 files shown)


ripienaar:
it's good, but unfortunately already on the list

aruntakkar:
But we should put it in the team collaboration Data or not?

ripienaar:
you could move it if you like, but I dont want it listed twice the list will become insane if we did that

aruntakkar:
Good no problem :)

PR #503, opened 2015-11-23
Per HP's announcement, HP Cloud will be terminated on January 31, 2016.

nclements:
Remove HP Cloud from list as they're killing it off.


ripienaar:
thank you

PR #507, opened 2015-11-25
Fix some links

evuez:
There was a missing "and" for "Design and UI" and "Tools for Teams and Collaboration".


ripienaar:
thanks :)

PR #510, opened 2015-11-29
Split out DNS into its own section

dandart:
Web hosting was a mix of web hosting and DNS - so this splits them apart

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Table of Contents
    * [PaaS](#paas)
    * [BaaS](#baas)
    * [Web Hosting](#web-hosting)
+   * [DNS](#dns)
    * [IaaS](#iaas)
    * [DBaaS](#dbaas)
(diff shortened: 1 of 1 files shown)


ripienaar:
nice, thanks

tomick:
@ripienaar Maybe move Cloudflare to DNS section?

dandart:
good point - it counts too

ripienaar:
I think they are a CDN first and the DNS stuff is a side feature. Not sure I am keen on listing the same thing many times - but then in their primary use first

dandart:
Oops - I duplicated above not realising that wasn't your intention - I'd personally have everything that does dns under dns etc, just for ease of search

ripienaar:
In the mid term I am going to write some web app for this data that has something like tags or a entry in different categories or such

This once I am done moving countries. Till then I am a bit weary of exploding the size of this rather inefficient current format

PR #524, opened 2016-01-03
Added Automated Browser Testing category, Gridlastic and also beanstalkapp

cheiff:
I leave the other services to @gavinengel, who suggested the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ Table of Contents
    * [Code Quality](#code-quality)
    * [Code Search and Browsing](#code-search-and-browsing)
    * [CI / CD](#ci--cd)
+   * [Automated Browser Testing](#automated-browser-testing)
    * [Security and PKI](#security-and-pki)
    * [Management Systems](#management-system)
    * [Log Management](#log-management)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,14 +134,17 @@ Table of Contents
   * https://snap-ci.com/ - Free for public repositories, 1 build at the time
   * http://www.appveyor.com/ - CD service for Windows, free for Open Source
   * [Comparison of Continuous Integration services](https://github.com/ligurio/Continuous-Integration-services)
-  * https://saucelabs.com/ - CI with scalable testing for mobile and web apps, free for Open Source
```
ripienaar: hmm, why move this one?

cheiff: That was the proposal of the OP.
Actually it fits in both categories. 
I can leave it as it was, if you want.
Another thing : what do I do with that problematic url? It has nothing to do with this PR.

ripienaar: right, yeah - i am not keen on dupes but we can put it wherever you and OP feel is best, you're right either will work

Ignore that link for now, I'll merge on red if URLs not from this PR is causing hassles, if it stays broken I'll remove it

PR #534, opened 2016-01-12
Add report-uri CSP and HPKP violation reporting

corburn:
If the server sets the Content-Security-Policy-Report-Only header, the browser will report violations of the content security policy to the given webhook. report-uri.io is a free service to log such violations.


ripienaar:
thank you

PR #541, opened 2016-01-27
Travis Update 

ReadmeCritic:
This pull request fixes issues with Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,4 @@ rvm:
 before_script:
   - gem install awesome_bot
 script:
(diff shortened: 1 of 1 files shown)


ripienaar:
so I am not sure, landscape.io is legitimately throwing a 500, I dont want to forget about it so i left it in - if it keeps throwing a 500 I will just remove it from the list rather than whitelist something that is actually broken.

sounds fair? what do you think?

TraderStf:
it's just their main page... a php bug or htaccess

ReadmeCritic:
Yeah feel free to close and just keep an eye on it :smile:

shakna-israel:
More on the landscape issue [here](https://github.com/landscapeio/landscape-issues/issues/209)

PR #543, opened 2016-01-29
Add houndify.com to List

tilomitra:
Houndify provides a nice API to add a conversational layer to any app or hardware, so I'm adding it to the list.


tilomitra:
Nvm, I see this has already been submitted in #543 so i'll close this.

PR #545, opened 2016-01-30
	Update Web Hosting - Removed Divshot.com

ChristianBagley:
Divshot.com was recently acquired by and is merging with [Firebase](firebase.com). They are shutting down their services and no longer accepting new accounts on [Divshot.com](divshot.com) and are encouraging new users to check out Firebase's new integrated hosting options (essentially what Divshot was offering). All in all, with the merge, Firebase is a great platform, but the addition of the service has made the subscription model that much more to digest.


ripienaar:
great, thanks for that

PR #546, opened 2016-02-02
Added fieldbook.com

jon-frankel:


ripienaar:
Nice. One I have been meaning to add as well :)

PR #548, opened 2016-02-02
Travis Update

ReadmeCritic:
White listing links with issues..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,6 @@ rvm:
 before_script:
   - gem install awesome_bot
 script:
-  - awesome_bot README.md --white-list github.com,jazz.net,yandex.com,debugmail.io --allow-redirect --set-timeout 5 --allow-timeout
+  - awesome_bot README.md --white-list github.com,jazz.net,yandex.com,debugmail.io,goreportcard,iron.io --allow-redirect --set-timeout 5 --allow-timeout
+
+  


ripienaar:
I really do not want to keep whitelisting sites thats just having some intermittent issue, I'd rather just disable the entire travis check if that's the only viable route.

Having these failures visible has value because I see over time the same site fails and just delete them from the list, whitelisting it just hides the bad apples.

ReadmeCritic:
Really depends how you want to use the tool

- here iron.io just is not working with travis so should be white listed imo
- goreportcard appears to be temporarily down.. but who knows
- white listing just helps you make the build `pass`, having the build pass allows you to pay attention to new issue when it fails (will you pay attention if the build is always failing?)

Anyway, as the owner you get to decide how you want things to work

Thank you for the list

ripienaar:
Problem is the last few weeks with this test have basically convinced me that expecting 100s of 3rd parties to be up for tests to pass is just a fools quest

So I don't expect this CI can ever be stable. 

Now it's useful because it helps me keep a manual eye on things but I think I should probably not run it via travis  at all it sends the wrong message to contributors and waste their time

I am probably going to remove it and run it instead somewhere it can give me the feedback of overall list sanity. 

The whitelisting of things that have issues are a mistake, we'll never know when they die. So I want to whitelist only for things like dupes for big players like AWS and not for small things.

PR #549, opened 2016-02-02
Remove Parse from list

j-delaney:
Unfortunately Parse is [no longer active](http://blog.parse.com/announcements/moving-on/).


j-delaney:
It looks like Travis is failing because of http://goreportcard.com/ and http://www.iron.io/

ripienaar:
thank you

PR #552, opened 2016-02-20
removed kimonolabs

shubhendusaurabh:
kimonolabs is shutting down its public service
https://kimonolabs.com/


ripienaar:
Thank you

PR #553, opened 2016-02-26
add inumbo.com

piccaso:


ripienaar:
Excellent. Once again discovering something I was personally looking for this week :)

TraderStf:
FYI, Cyren offers a 'filtering-forwarding' which is very effective.
But that also means all your emails are passing through their servers.

myusualemail@...  forward all to  myemailforcyren@... forward and filter to your inbox

http://www.cyren.com/cyren-emailsecurity.html

PR #554, opened 2016-02-26
remove mandrill.com

piccaso:
free plan discontinued on Feb 24, 2016
http://blog.mailchimp.com/important-changes-to-mandrill/

Quickfix if you need an alternative and use SMTP: http://mailgun.com/


ripienaar:
Thank you

PR #559, opened 2016-03-01
Update Algorithmia to reflect the new free tier

anowell:
No longer just a batch of upfront free credits. Now includes free monthly credits.

http://blog.algorithmia.com/2016/03/algorithmia-is-now-free-forever/


ripienaar:
Nice, thanks :)

PR #565, opened 2016-03-16
qtcloudservices seems gone

zoidbergwill:
Opening the URL or the base domain says 

> No Application Configured
>
> This domain is not associated with an application."


ripienaar:
Thank you

PR #567, opened 2016-04-07
Add TinyLetter.com to [Email]

NurdTurd:
I've added TinyLetter (tinyletter.com) to the 'Email' section. I could not figure out how many emails you could send per month for free, so I only listed the subscribers.


ripienaar:
thanks :+1:

NurdTurd:
No problem, I love what your doing here. :+1:

PR #568, opened 2016-04-07
Added biz.mail.ru DNS + email

SpartakusMd:
biz.mail.ru free DNS & email + cloud for 5000 users per domain with 25 GB space each


ripienaar:
I prefer to add them to just one section - i know in this case it's a bit tough to pick one, but if we start duplicating them the various IaaS providers would end up in almost all cats

SpartakusMd:
@ripienaar, I thought about this and saw `domain.yandex.com` which is duplicated so I decided to do it. In this case I will remove the entry from DNS, because biz.mail.ru is more about emails :)

ripienaar:
Indeed there are some dupes, gifts fix them one day. Not adding more helps :)

ripienaar:
Thanks !

PR #570, opened 2016-04-10
Added Assembla back in

Reubend:
I think that Assembla may have been mistakenly removed. The old text listed a 7 day trial which appears to be for their task management tools, not their repos.  The page I linked to claims that the repos are still free and doesn't list a time limit. If this isn't suitable, could someone provide the reason why?


ripienaar:
Yeah you're right repos seems free free. Thanks

PR #571, opened 2016-04-12
Added new services, deleted obsolete ones, and reorganized sections

Reubend:
Put API and data services into their own section
Put BaaS services which were listed as PaaS services into the correct section
Added Domino Data Lab
Added Haven
Removed Helion, which no longer offers public cloud services
Merged Docker categories since there are only 2 services there
Merged Vagrant categories since there are only 2 services there
Removed Codio, which does not appear to offer any free plan anymore
Moved NS1 from IaaS to DNS because it is a "Data Driven DNS" 
Added a miscellaneous section
Moved Formlets, Superfeedr, and Instadiff out of PaaS to misc because none of those are a PaaS
Moved Readme.io to misc because it is not a database, it is documentation hosting

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,8 +483,6 @@ Table of Contents
    * [apiary.io](https://apiary.io/) — Collaborative design API with instant API mock and generated documentation (Free for unlimited API blueprints and unlimited user with one admin account and hosted documentation)
    * [mockable.io](https://www.mockable.io/) — Mockable is a simple configurable service to mock out RESTful API or SOAP web-services. This online service allows you to quickly define REST API or SOAP endpoints and have them return JSON or XML data
    * [jetbrains.com](https://jetbrains.com/products.html) — Productivity tools, IDEs and deploy tools. Free license for students, teachers, Open Source, and user groups
-   * [readme.io](https://readme.io/) — Beautiful documentations made easy, free for Open Source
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -347,6 +333,7 @@ Table of Contents
   * [zapier.com](https://zapier.com/) — Connect the apps you use, to automate tasks. 5 zaps, every 15 min. and 100 tasks/month
   * [stackstorm.com](https://stackstorm.com/) — Event-driven automation for apps, services and workflows, free without flow, access control, LDAP,...
   * [simperium.com](https://simperium.com/) — Move data everywhere instantly and automatically, multi-platform, unlimited sending and storage of structured data, max. 2,500 users/month
+  * [stamplay.com](https://stamplay.com/) — 50 K API calls, 100 GB data transfer, and 1 GB storage for free
```
ripienaar: can you add a small description about what this is in addition to the limits? thanks!

ripienaar:
looks great thanks! if you want to fix that let me know when done else just ping here if not, it's good to merge

Reubend:
The original PR was just to move Stamplay to the correct section (because it's not actually a PaaS) but I have added a more descriptive explanation now. BTW, thanks for the great work maintaining this repo!

ripienaar:
yes indeed thanks @Reubend I noticed that later.  thanks a ton for this

PR #579, opened 2016-04-22
Add now on PaaS section

lius:
now is a new PaaS hosting option for Node.js applications, featuring auto scaling by default (https://zeit.co/now)


ripienaar:
thnx

PR #582, opened 2016-05-08
Add Stackdriver Error Reporting in the list of exception monitoring

steren:


ripienaar:
The pricing here is not clear. It says "try it free" but where does it show price details? 

Am looking from a phone maybe it's just not visible here but worth double checking 

Thanks

steren:
You are right that: the feature is entirely free while beta but final prices have not been communicated.

ripienaar:
Hmm. Better to add it once pricing is clear then I think. Adding such services leads to bitrot sooner rather than later

So I will close this one but appreciate the effort in making a PR

PR #586, opened 2016-05-14
Remove assembla.com

### LINKED ISSUE #585: Assembla is only 7 days of free trial
Unless I missed something, Assembla is now only 7 days of free trial. May be it should be removed from the list.

jeromerobert:
* Now only 7 days of free trial
* Close #585


ripienaar:
thank you!

PR #590, opened 2016-05-22
Adding Backblaze B2

DevWurm:
Adding Backblaze B2 cloud storage.
Free 10GB for unlimited time with some restrictions regarding access volume (but not extremely limiting)


ripienaar:
Thank you

PR #593, opened 2016-05-24
scinote.net

SangramChavan:
scientific data management & team collaboration. One Team with Unlimited number of users, backup and 1GB storage space


ripienaar:
nice, bit on the fence of what I think the target audience of the list will find useful, but merging anyway :)

thanks!

PR #595, opened 2016-05-25
Deploybot no longer have a free plan

danzel:
If they do I can't find it.
Their sign up page doesn't mention it https://signup.deploybot.com/signup/new


ripienaar:
Thanks, that's how it goes :)

PR #598, opened 2016-06-04
updating deferpanic.net link

eyberg:


ripienaar:
if it's just free during beta and not free later on it wont be a good fit for this list, its for services with a free tier for ever, do you know if they plan to offer such a service after beta too?

eyberg:
yes, there will def. be a free tier after the beta, I think it's well with inline w/the aforementioned rackspace/aws plans (their free tiers are valid only for 12 months) as well

ripienaar:
I can only go by what's on their site and its pricing page and to me it doesn't seem to feature a free tier. Free short trial yes but those don't fit here

Link details here confirming their free plan ideas and I will gladly merge it

PR #600, opened 2016-06-08
Warn about x10hosting's country restriction

lubien:
> Sorry, we're not able to accept your registration.
> We are not currently accepting new users from Brazil. Please try again at a later date. We are expanding our allowed country list in the upcoming days.

It's an old issue for some countries. See [this](https://x10hosting.com/community/threads/x10hosting-country-block-just-remove-it.164342/). 

Maybe it would be good to add


ripienaar:
Oh that's bullshit. They don't belong on the list then. Let's just remove them.

ripienaar:
I'll revert it once at a computer again. Thanks for catching that. Closing this one

PR #608, opened 2016-06-19
http://www.baasbox.com/

SangramChavan:
The open source backend for your mobile app

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,7 @@ Table of Contents
 
 ## BaaS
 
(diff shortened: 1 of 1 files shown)


ripienaar:
Thanks for the contribution. Do they offer some hosted version of this? This list is not for self host software but more stuff others manage. If we started listing self hosted too we would immediately have thousands of entries so gotta focus unfortunately.

SangramChavan:
No they don't provide hosting version.

PR #610, opened 2016-06-27
Mailinator.com

SangramChavan:
Mailinator is Free, Public, Email system where you can use ANY inbox you want! ... Disposable Email.
Plans - https://www.mailinator.com/featurematrix.jsp


ripienaar:
Thanks

PR #613, opened 2016-06-30
Removed airbrake.io

SangramChavan:
No more Free Plan. Only a 30 days Try for free period

Thanks to @Adrion !!


ripienaar:
thanks both for sorting that out

PR #616, opened 2016-07-07
added docker.com

SangramChavan:
* [docker.com](https://www.docker.com) — One free private repository,free managed node and Unlimited public repositories
Plan - https://www.docker.com/pricing


ripienaar:
weird! wonder how that never made the list :) thanks

PR #620, opened 2016-08-01
Removing SourceLair

edgardmota:
SourceLair doesn't offer a free forever plan anymore as stated at their blog: https://www.sourcelair.com/blog/articles/100/free-becomes-basic.


ripienaar:
👍

PR #621, opened 2016-08-03
gitlab.io

mradzinski:
Added gitlab.io, a free git source code repo


ripienaar:
this one is unfortunately already here, see `about.gitlab.com`

thanks for the effort though really appreciate it 👍

PR #631, opened 2016-08-23
teams: Added meet.jit.si

sbadia:
Hi here, just a litle pr in order to add jitsi in the teams section.
Jitsi can be self-hosted, but has an SaaS offer.

* video conversations
* screen sharing
* unlimited users

Thanks!


ripienaar:
thanks 👍

PR #633, opened 2016-08-27
rm code.fosshub.com - permclosed

bithakr:
Source: [Official Statement by FossHub_com on Reddit, linked from mainpage](https://www.reddit.com/r/sysadmin/comments/4vzovk/fosshub_statement_regarding_2nd_august_security/).

_"As a response to this, we will be temporary shut down OldFoss to clean this, our repository for older versions and we decided to **close down permanently Code.FossHub.com** a free service that we offered hoping it will help some free projects after Google Code was deprecated. We will not abandon the existing, legit users who still use it and will continue to offer them the same service."_


ripienaar:
Thank you

PR #639, opened 2016-09-25
Added MarkLogic (Enterprise NoSQL Database Platform)

wlsc:
+ added MarkLogic (Enterprise NoSQL Database Platform), which has free developer license.


wlsc:
Maybe it is even better to open "Database" section and move it to that category?

ripienaar:
Seems to me the Devi version is a download? This list is for *aaS i.e. Hosted and offered as a service and where that service has a free tier.

wlsc:
@ripienaar they have templates to build backend automatically in AWS. You search "marklogic" keyword in the AWS Marketplace and then deploy it into your AWS instance for free (with the developer license).

More information you can find here: http://developer.marklogic.com/products/aws

ripienaar:
Even in that case you are still effectively maintaining self hosted software. Any software can be installed into AWS. That does not make it a system someone else maintains, backups etc 

Sorry, I do t think it's a good fit for this list

PR #646, opened 2016-10-03
Adding Crashlytics to Crash Handling

viswamitra:
Crash Reporting made easy with Crashlytics. Free for iOS and Android.


ripienaar:
this one has been contributed a few times and I've not merged them because this list tend to be slanted towards `things infrastructure developers (System Administrator, DevOps Practitioners, etc.) are likely to find useful` (from the README) and I feel this one is a bit niche to fit that bill

What do you think?

viswamitra:
If the list is strictly devops focused, then yes, I agree. But I have seen some sections like 'IDE and Code Editing' which are pretty much developer focused as well. Tools and Services that lets you monitor and measure the performance would also fall under devops?

ripienaar:
So the list is for *developers* who are infrastructures focused, so IDEs and such are a very good fit, especially with the popularity of infrastructure as code - we do tend to spend a very large amount of time in IDEs!

infra devs are those focussed on tooling internally, open source tools in that space etc, I just dont find a monitoring tool thats specifically mobile orientated generic enough

So crash reporting for general apps good, crash reporting for limited platforms bad, basically

viswamitra:
Alright! I have one more list item - [Splunk](https://www.splunk.com/). There seems to be a Free section, where you have various core products offered. Would this be a good fit?

ripienaar:
Splunk would be a good fit but I've never been able to find out what their hosted product offers in the way of free - like it says 'trial' but does it mean a few days? or limited by volume but perpetual, or perhaps a particularly long free period like AWS being a year

Only their splunk cloud thing would really fit here - if you can find a link to whats free and its not say a 30 day trial it'd be great

viswamitra:
[This](http://docs.splunk.com/Documentation/Splunk/6.5.0/Admin/MoreaboutSplunkFree#Is_Splunk_Free_for_you.3F) link says that the Splunk Free version lets the user index 500 MB per day and it doesn't expire. 
Looks like that will fit the bill?

ripienaar:
Yeah but that's self hosted. List is for hosted services like their cloud offering wild be

PR #651, opened 2016-10-04
Remove mailstache.io

ericthornton:
Mailstache's service is ending so obviously no more free plan.

>Mailstache will be ending services on November 1, 2016 at 12:00 AM MST.
All account data and email will be completely deleted and disks will be
wiped clean.


ripienaar:
thanks! 👍

PR #652, opened 2016-10-04
Codefresh - Continuous Delivery and Collaboration Platform for Docker uses

dockermaster:
… users


ripienaar:
nice one, thank you

PR #654, opened 2016-10-12
Added KeyCDN

punit2502:


ripienaar:
unfortunately this is more free trial than free service for development purposes so I cant accept this one sorry, thanks for contributing though!

PR #662, opened 2016-11-01
added buddy.works to repository services

luord:
They offer a free private project to every new account, with a git repository (or integration for other git services) and basic pipelines for continuos delivery.


ripienaar:
👍  thanks

PR #677, opened 2016-11-20
Removed PubStorm.com

SmartFinn:
The [PubStorm.com](http://www.pubstorm.com/) hosting service was shutdown on November 18th, 2016.


ripienaar:
Thank you

PR #680, opened 2016-11-24
Add TryStack - Free Openstack hosting

unixfox:
Free Openstack hosting, suitable for testing only because they reset the environment every 24 hours.


ripienaar:
neat

ripienaar:
👍

PR #682, opened 2016-12-09
Remove OpsGenie

EvanKrall:
Opsgenie got rid of their free tier. It looks like existing accounts are grandfathered in at the Lite level.


ripienaar:
Thanks!

PR #694, opened 2017-01-17
Remove facetflow

cassiosantos:
It is no longer available: https://twitter.com/facetflow/status/742338350581960704


ripienaar:
thanks for the update 👍

PR #704, opened 2017-02-09
renamed mongolab.com to mlab.com

mislavcimpersak:
In February 2016 MongoLab.com changed it's name to mLab.com

source:
https://techcrunch.com/2016/02/29/mongolab-is-now-mlab/


ripienaar:
thank you

Issue #707, opened 2017-02-16
Openshift has changed pricing


agauniyal:
Under the new name of Next Gen, free accounts are limited to 1 month trial only.

ripienaar:
shitty, feel like making a PR to kill it off the list? we dont do short free trials on this list

agauniyal:
Sure - https://github.com/ripienaar/free-for-dev/pull/708

PR #708, opened 2017-02-16
Remove Openshift from the list

agauniyal:


ripienaar:
thanks 👍 

Closes #707

PR #712, opened 2017-03-07
Added the Atom Editor

neikei:


ripienaar:
thanks for the contribution, nothing against atom - but we don't tend to list downloadable, desktop stuff, self hosted apps etc

sorry but thanks for the effort

PR #713, opened 2017-03-07
Added Github Pages

neikei:


ripienaar:
again really appreciate the contribution, but we should rather flesh out the existing github description - can you imagine if we were to try and add everything AWS and GCE does individually?

PR #723, opened 2017-04-12
Add Figma

devmattrick:


devmattrick:
@ripienaar its in a bit of an odd place at the moment; its completely free (all features are free with no restrictions) until July 1st. Should I just update it to reflect the changes that will take place on July 1st?

ripienaar:
Yeah I think that's best else we forget :P in general early stage ones I only take if its clear what the long term plan is and then I stick their long term plans up thats safest

ripienaar:
Thank you @devmattrick

PR #727, opened 2017-04-23
codota.com

SangramChavan:
[codota.com](https://www.codota.com/) — Codota helps developers create better software, faster by providing insights learned from all the code in the world. Plugin available.


ripienaar:
thanks

PR #729, opened 2017-04-23
Fixing CARTO Url

facundofarias:


ripienaar:
thank you @facundofarias 👍

PR #731, opened 2017-05-06
Added Sublime Text to IDE and Code Editing section

Marketionist:
**This pull request adds Sublime Text to IDE and Code Editing section**


ripienaar:
This list is for hosted services with free tiers. Not for self hosted apps or editors

Now I know we have a few editors here but I think I need to draw the line and rather delete the ones listed already. It's just not in line with the lists goal and in hindsight I think I accepted them in the past by mistake

PR #733, opened 2017-05-20
Add Mattermost as team collaboration tool

Rugal:
My company is using it, quite good.


ripienaar:
This is self hosted though isn't it @Rugal ? This list is for services offered in the *aaS model, so self hosted wont count

Rugal:
@ripienaar Oh you are right, it provides only self-hosted version. 
Will close this PR.

PR #736, opened 2017-05-30
Remove Codepunker CDN

antimatter96:
Directly quoting the website

```
This tool has been removed because of multiple flooding attempts and it will not be re-enabled.
Those who used it legitimately until now may continue to use it until April, 20th 2017 
when all assets will be deleted without any other notice.

We apologize to all those who have been harmed in any way.
```


ripienaar:
lol, thanks for that!

PR #738, opened 2017-06-12
Added Amazon DynamoDB to DBaaS section

Darkle:


Darkle:
Yeah i guess that makes sense. The reason i wanted to add DynamoDB specifically is because I don't think it's well known that AWS has a DBaaS. I think most people (including me till recently) thought all their databse services ran on spun up servers. Also another reason I thought DynamoDB was worth adding was because I was looking for a NoSQL DBaaS and while mLab is pretty good, it doesn't backup/replicate the free tier database, whereas the DyanamoDB service does.

ripienaar:
Makes sense. Though I think every AWS service have unique / novel features. There are lists for that - this one I think I prefer to not list the same product group too much.  Some exceptions here already to that but I'd rather remote then then add more. 

Appreciate the effort in opening the PR but going to have to be a no from me sorry

PR #748, opened 2017-07-25
ArvanCloud CDN was added

aligholamee:


JABirchall:
Do they have a free CDN?
I looked at their pricing and it says they only provide free DNS the CDN is paid, and poorly priced as well.
Either;

- $0.90 for 15gb/m
- $90 for 1500gb/m

Nothing in-between, terrible pricing model.

ripienaar:
yeah I don't see it either, closing it, if you have a link to their free info please post here and reopen

PR #749, opened 2017-07-27
outdated services

untoreh:
- iriscouch was acquired dead domain
- elastx despite still advertising a free service does not allow envs after account creation, seems free acounts request are manual
- cloudandheat no traces of free offerings on their website


ripienaar:
Awesome thank you

PR #755, opened 2017-08-21
Adds Session Recording, split from Analytics.

jagracey:
- Adds getWisdom.io
 - Adds UserSurge.com
 - Pulls down Inspectlet, Hotjar, MouseStats.
 - Updates Table of Contents.


ripienaar:
nice one! thanks :)

PR #763, opened 2017-09-07
Removing imagefly.io, which has shut down

bengadbois:
> "We have unfortunately decided to discontinue our Imagefly offering. Our company has gone through a transition to other projects, and we find ourselves unable to invest the resources needed to provide Imagefly customers with the level of service and support that they deserve."


ripienaar:
Thanks for the update!

PR #767, opened 2017-09-11
add prismic

nurazhar:
headless CMS


ripienaar:
nice, we can probably do a bit better with the description, for example they have perpetual for for open source package, details of that would be good here

bloep:
is it better to add this to a section called CaaS (Content as a service), isn't it?

ripienaar:
closing, lack of response

PR #768, opened 2017-09-11
Update README.md

bravo-git:
Added electron framework in the Design and UI section.


ripienaar:
This list is for Software As A Service, this seems like a library so unfortunately will not qualify

bravo-git:
We have IDEs listed as a SaaS in this list. Sorry, I can't understand why a FOSS framework  can't qualify for SaaS which is used in developing desktop apps .  Note that the term SaaS is not rigorously defined( source: [Wiki](https://en.wikipedia.org/wiki/Software_as_a_service)).
According to your list free online text/image editors, analytics tools, pastebins can qualify in SaaS category if you wish I can add more to above mentioned things. Time is not that far away when there will be minimal difference between FOSS and SaaS.

ripienaar:
That's true and those were exceptions too. As you say it's not well defined but this is not a list for software libraries or frameworks

Sorry I know it's not well defined or scientific. I do now what I made the list for though and I do not think listing tons of frameworks and libraries in this list will be helpful. There are many "awesome lists" covering many fields and groups there are no doubt a more appropriate place for this listing. Sorry.

PR #770, opened 2017-09-12
Add Gravit Designer to Design and UI

AAlakkad:


ripienaar:
Thanks for the contribution. Unfortunately this list focuses on SaaS offerings. We have the odd IDE but in general it's a SaaS list so this won't fit 

Thanks.

PR #771, opened 2017-09-12
Removed placepenguin

OrBin:
[placepenguin.com](http://placepenguin.com/) is not a service for placeholder images of penguins any more.


ripienaar:
Thank you

PR #776, opened 2017-09-14
Add WeDeploy service

davidalves1:
Add [WeDeploy](https://wedeploy.com/) service in web hosting session


davidalves1:
The [Terms of Service](https://wedeploy.com/terms/) just says:

**7. Fees**
There are currently no fees associated with an End Users use of WeDeploy during the Beta Term. Use of WeDeploy may be subject to fees under a separate agreement upon conclusion of the Beta Term.

ripienaar:
OK then I'd rather not add it sorry, else it just becomes false info later.  Happy to take beta products that for sure say they will have a free tier later

thanks for the effort though!

davidalves1:
No problem. After the beta is finished, if it has a free level, I'll open the pull request again

PR #777, opened 2017-09-21
Update info for Pusher

minaorangina:
Some out-of-date info, plus some new stuff


README.md
```diff
@@ -427,7 +428,7 @@ Table of Contents
 ## STUN, WebRTC, Web Socket Servers and Other Routers
 
    * [scaledrone.com](https://www.scaledrone.com/) — Push messaging service. Free for up to 20 simultaneous connections and 100,000 messages/day
-   * [pusher.com](https://pusher.com/) — Hosted Web Sockets broker. Free for up to 20 simultaneous connections and 100,000 messages/day
+   * [pusher.com](https://pusher.com/) — Realtime messaging service. Free for up to 100 simultaneous connections and 200,000 messages/day
```
ripienaar: prefer we just list it once

minaorangina: Can combine into one. Issue is those two products belong to separate categories.  And both happen to be free.  Thoughts?

ripienaar: Ah I see, its 2 entirely seperate things, fair, will merge as is, thank you!

PR #778, opened 2017-09-21
Remove DNS Spy, there is no longer a free tier

mattiasgeniar:


ripienaar:
👍 but I think you need to rebase or something

mattiasgeniar:
Woops, updated the PR!

PR #781, opened 2017-10-01
Added "Bitbucket Pipelines"

bloep:


JABirchall:
BitBucket pipelines is a service only avalble if your using bitbucket right? Its not something you can use if your using github for instance.

bloep:
Yes, it's Bitbucket only.

ripienaar:
bitbucket is already listed, I'd rather if there's some significant addition we just updated the existing description

ripienaar:
👍 thank you

PR #785, opened 2017-10-12
add teachablemachine.withgoogle.com into ML section

EternalReign:
add https://teachablemachine.withgoogle.com into ML section


ripienaar:
I don't think this really is a software as a service, its handy and useful, but does not really fit the scope of this list I think,

Sorry, really appreciate the effort!

PR #787, opened 2017-10-13
Add Github Pages

juniorRubyist:
Add Github Pages


ripienaar:
it's relady mentioned (github) so I think we can update one of the other descriptions, things will get unwieldy if we keep adding these big multi feature services multiple time? Generally I say no to these but I also see its already listed a few times, not really sure

ripienaar:
closing, lack of response

PR #799, opened 2017-10-24
Walkme doesn't redirect to www

CorralPeltzer:


ripienaar:
weird? the old one worked fine for me, but yeah might as well update it to what the final address is

thanks 👍

CorralPeltzer:
Yes, my bad, the site redirects correctly, I had a problem with a wrongly configured local proxy and the presence of `walkme.com` in [this blocklist](https://github.com/mitchellkrogza/Badd-Boyz-Hosts).

PR #807, opened 2017-11-01
Add Riot.im

caffeinewriter:
### No limits on features, group size or usage for public and private rooms

https://about.riot.im/riot-is-free/


ripienaar:
thanks 👍

PR #809, opened 2017-11-09
Added Hasura

praveenweb:
Have added Hasura.io - PaaS + BaaS solution to build apps fast. 
Hasura provisions a managed server with BaaS APIs for Data, Auth and Files. 
It provides complete access to underlying infrastructure;
Free for single node cluster. Pricing plans for upgraded infrastructure + services.


ripienaar:
👍 thank you

PR #812, opened 2017-11-29
corrected txti.es url

reed-jones:
url had missing "i" and https doesn't work


ripienaar:
doh I did test this one but I think I have this bad habit of copying the description url and not the actual one so it slipped through!

thanks for the fix!

reed-jones:
No problem, Thanks for the list :)

PR #815, opened 2017-12-05
Added plan.io

bloep:
Added plan.io
> We also offer the free plan Bronze: 1 project, 2 users, 10 customers, and 500 MB storage

See: https://plan.io/pricing/


ripienaar:
thank you :)

PR #819, opened 2018-01-11
Add kanbanize.com

paffas:
Hi. It will be great if you can add Kanbanize.com 
A kanban software that is suitable for managing multiple projects. There is a 30-day free trial. 
I can see that not tools are totally free, such as Leankit and Kanbanery. If you have some questions do not hesitate to ask. :-) 
Regards.


ripienaar:
@paffas those two did offer a free offering at a time, they must have changed what they offer.  Trials are not accepted sorry

PR #824, opened 2018-01-16
New line for "Have I been pwned?" 

baknu:
With my previous PR I accidentaly deleted the enter before "Have I been pwned?"
Sorry and thanks for fixing.


ripienaar:
ah, i missread the PR my bad, thanks for the fix

PR #828, opened 2018-02-02
Removed filehero

shakna-israel:
As per comments in [424](https://github.com/ripienaar/free-for-dev/pull/424#issuecomment-361814626), filehero is dead.


ripienaar:
thank you :)

PR #834, opened 2018-02-19
Add LogRocket (logrocket.com) entry

begincalendar:


ripienaar:
hmm, handy and free yes, but not sure it's in line with the scope of the list that I am aiming at :(

begincalendar:
Can you please elaborate on how it does not align with the scope?

ripienaar:
`The scope of this particular list is limited to things infrastructure developers (System Administrator, DevOps Practitioners, etc.) are likely to find useful. We love all the free services out there, but it would be good to keep it on topic. It's a bit of a grey line at times so this is a bit opinionated; do not be offended if I do not accept your contribution.`

I just dont think devops types might find this useful in the tools THEY build.  I see how its useful in the general case though.

PR #836, opened 2018-02-23
Postmark Price changes

kevin51jiang:
No longer 25 000 emails free, just a dev account for 100 free/month.

https://postmarkapp.com/pricing


ripienaar:
for a 100 emails I think we should just remove them from the list, cos whats the point of that.

Want to kill it?

beesofts:
yes it can be simply removed

kevin51jiang:
Postmark Removed

ripienaar:
thank you

PR #837, opened 2018-02-23
Removed Atlassians Jira because of 404

KloseD:
The link ends in a 404. As far as I can see, there's only a 30 Day Trial of Jira.


ripienaar:
they still do open source but its so hidden you will never find it

https://www.atlassian.com/software/views/open-source-license-request

so I am happy to remove it if they cant be bothered making it clear

ripienaar:
I routinely reject services that makes it near impossible to verify they offer a free service. 

It’s just too much effort to chase it down and so forth if it’s not clear it’s not on the list I am afraid.

PR #847, opened 2018-03-27
Fix link for glitch

unixfox:
The right domain is `glitch.com` because the domain `www.glitch.com` doesn't exist.


ripienaar:
It does exist, they just appear to have a broken config - anyway, fixed thank you!

PR #849, opened 2018-04-04
migadu.com

piccaso:
proposing migadu.com email hosting:
limited outgoing mails but still useful for receiving. 
even supports subdomains.


ripienaar:
Thank you

PR #851, opened 2018-04-19
 Removed backand.com since they are not in service anymore

jrobinsonc:
Backand.com closed its service on February 28, 2018


ripienaar:
Thank you!

PR #853, opened 2018-04-23
Add bintray.com for artifact repos

amdelamar:
They have free hosting for oss project artifacts in maven, npm, docker, and many more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Table of Contents
 
 ## Artifact Repos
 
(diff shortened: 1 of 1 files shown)


ripienaar:
thankI merged this and then reverted it - bintray is there already perhaps just in a dumb place, maybe the better PR would be to move it to the artifact store section from where its now?

amdelamar:
Oh it is? Whoops.
Yeah that makes sense to move it to Artifact Repos because of its usage. Let me know if you want another PR.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ripienaar:
if you have time another would be great yeah, thanks @amdelamar

Issue #866, opened 2018-05-20
Add no-ip 


gerardo-junior:
a dynamic dns service that allows up to 3 hostnames and limited domains free with (with confirmation every 30 days)

https://noip.com

PR #867, opened 2018-05-20
Added noip

gerardo-junior:
a dynamic dns service that allows up to 3 hostnames free with confirmation every 30 days

#866


ripienaar:
thank you

Issue #868, opened 2018-05-20
Add play-with-docker


gerardo-junior:
A simple, interactive and fun playground to learn Docker

https://play-with-docker.com/

ripienaar:
thank you!

PR #870, opened 2018-05-21
Fixing url play-with-docker

gerardo-junior:
It seems that the url of previous PR is not doing redirect no more, I'm sorry.

#868 #869


ripienaar:
Thanks :)

PR #879, opened 2018-07-11
Removed abandoned IaaS items

rvben:
virtzone.net and dply.co are down, and seem abandoned.

dply.co - Last Twitter message on Jan 16, and last repo update in February.
virtzone.net - No signs of life


ripienaar:
awesome thank you

PR #889, opened 2018-08-20
Remove ServerPilot.io

codyogden:
As of July 10, 2018, [ServerPilot no longer offers a free version](https://serverpilot.io/blog/2018/06/08/pricing-changes). It should be removed from this list.


ripienaar:
thanks!

PR #890, opened 2018-08-23
Add www.checkbot.io

seanw:
I'm the creator of www.checkbot.io. Hopefully you'll find it relevant. Thanks for the great resource!


ripienaar:
generally i wouldnt take a browser extension but this looks like it might fit :)

ripienaar:
thanks

PR #894, opened 2018-08-27
Remove Request Bin

clarmso:
RequestBin is no longer online, but its github repo is available.  The instructions are available to host the local version of request bin.

https://github.com/Runscope/requestbin


ripienaar:
thank you :)

PR #897, opened 2018-08-28
Okta acquired Stormpath

clarmso:
Stormpath has shut down its API in 2017.  It has been acquired by Okta, which offers a free tier.

See https://stormpath.com/blog/stormpaths-new-path


ripienaar:
thank you :)

PR #899, opened 2018-08-30
Remove Gemnasium.com

jarikmarwede:
Gemnasium.com does not exist anymore because their team got acquired by Gitlab as seen when visiting [their website](https://gemnasium.com)


ripienaar:
thank!

PR #902, opened 2018-08-30
Remove dependencyci.com

jarikmarwede:
dependencyci.com does not seem to exist anymore because it redirects to [Tidelift](https://tidelift.com/)


ripienaar:
thank you!

PR #908, opened 2018-09-01
Remove athenalayer.com

jarikmarwede:
[Athenalayer.com](http://athenalayer.com/) redirects to [OrcaTech](https://orca.tech/web-tools/) which means that it probably does not exist anymore.


ripienaar:
Thanks for all these edits, so awesome !

jarikmarwede:
No problem, I'm just going through it myself and fixing some stuff that doesn't seem right :)

PR #920, opened 2018-09-02
Rename placehold.it to placeholder.com

jarikmarwede:
[Placehold.it](https://placehold.it) got renamed and now redirects to [placeholder.com](https://placeholder.com)


ripienaar:
👍

PR #922, opened 2018-09-02
Remove taxratesapi.avalara.com

jarikmarwede:
[taxratesapi.avalara.com](http://taxratesapi.avalara.com/) redirects to https://developer.avalara.com/avatax/signup/ which seems to only have a free trial as seen on the [Overview page](https://developer.avalara.com/avatax/).


ripienaar:
👍

PR #923, opened 2018-09-02
Rename reappt.io to pushtechnology.com

jarikmarwede:
[Reappt.io](https://reappt.io) seems to have been renamed to [Pushtechnology.com](https://pushtechnology.com).
The pricing is still the same.


ripienaar:
👍

PR #946, opened 2018-09-02
Remove appacitive.com

jarikmarwede:
[Appacitive.com](https://appacitive.com) no longer exists.


ripienaar:
awesome, I sent you an invite if you want to be a colaborator to save you some hassle on the PR dance!

jarikmarwede:
Also thanks for replying so fast all the time

Issue #961, opened 2018-09-13
Remove newrelic.com


bazzilio:
Catalog said:
`newrelic.com — Free with 24 hours data retention`
in realty:
![image](https://user-images.githubusercontent.com/13622424/45518488-5f91f080-b7ba-11e8-8ffa-aeb65b9f3c8d.png)

We should remove newrelic.com  from this catalog

ripienaar:
Used to yeah. Alas. Want to send a PR to remove?

bazzilio:
> Used to yeah. Alas. Want to send a PR to remove?

i would try.
It would be my first PR :)

bazzilio:
Some people may be interested in free APM solution for newrelic

![image](https://user-images.githubusercontent.com/13622424/45520100-805d4480-b7c0-11e8-8988-88c5ae3d6e69.png)

ripienaar:
house keeping ancient issues

PR #962, opened 2018-09-13
Issue #961 Remove newrelic.com

bazzilio:
Seems, like newrelic did not provide server monitoring for free.
As i found in #961 it provide very limited APM for 2 hours.


ripienaar:
Trial became lite if you didn’t activate pro. And that kept data for 2 hours. That’s how it works. Which is fine for the charter of this list.

ripienaar:
Thank you :)

bazzilio:
> Trial became lite if you didn’t activate pro. And that kept data for 2 hours. That’s how it works. Which is fine for the charter of this list.

probably, it would be good to split monitoring from APM.

PR #973, opened 2018-09-20
Added Galiboo's Music API (galiboo.com)

thecooltechguy:


README.md
```diff
@@ -604,6 +604,7 @@ Table of Contents
 ## APIs, Data and ML
 
   * [Scraper API](https://www.scraperapi.com/) — Cloud based web scraping API handles proxies, browsers, and CAPTCHAs. Scrape any web page with a simple API call. Get started with 1000 free API calls/month.
+  * [Galiboo Music API](https://galiboo.com/) — An AI-powered API for music.
```
ripienaar: `Access to our demo catalog of ~20K tracks` this seems to be too limited to be of much use, also doesn't really fit at all in with the scope of tools we gather here.

So afraid I can't accept this one, thanks a lot for the effort though!

thecooltechguy: Oh ok, no worries! :)

PR #975, opened 2018-09-20
Added Rocket.chat - Replace email, HipChat & Slack

kamilzadora:


ripienaar:
I am afraid we dont include self hosted software in the list, the hosted option has to have the free tier

thanks a lot for the effort though!

PR #783, opened 2017-10-06
Remove sendloop

ericthornton:
[Sendloop's pricing page](https://sendloop.com/pricing) indicates that they do not have a free plan. 

> Unfortunately we don't have a free plan. Building and maintaining a world class email delivery infrastructure costs a lot.

PR #984, opened 2018-10-02
Remove sendloop...again

ericthornton:
I removed sendloop around this time last year, with #783. It was added back in #788 referencing a forever free plan. The https://start.sendloop.com now redirects to just https://sendloop.com and there is this bit on their [pricing page](https://sendloop.com/pricing/):

> Can I use Sendloop for free?
Unfortunately we don’t have a free plan. Building and maintaining a world class email delivery infrastructure costs a lot. Sendloop has one of the biggest email delivery platform and the small amount of fee we charge to you helps us a lot to keep this platform active.


ripienaar:
Thank you

PR #987, opened 2018-10-07
Added the 1984.is Free DNS service

dragstor:
The 1984.is offers Free DNS services with zones, slave zones, API and other free stuff included.


ripienaar:
thanks

PR #995, opened 2018-10-29
Fastly no longer offers free $50/month

cjd118:


ripienaar:
Thanks for the update :)

PR #997, opened 2018-11-06
Add humio.com

cornet:
Hosted log management with free tier of 2GB/day.

Details of free tier can be found here https://www.humio.com/pricing


ripienaar:
thank you

PR #1004, opened 2018-11-26
Remove DBaaS/orchestrate.io

OleksiyRudenko:
The service doesn't respond.

According to this [post](https://news.ycombinator.com/item?id=12936031) on ycombinator the service is retired since March 31, 2017 (announced date was December 15, 2016 and then extended).


ripienaar:
thank you @OleksiyRudenko

PR #1007, opened 2018-11-30
Add free retrospective tool

oyaaraas:
https://www.senseitool.com - Fun, effective retrospectives for distributed agile teams.


ripienaar:
thank you

PR #1021, opened 2019-01-30
Update README.md

prathaprathod:


ripienaar:
while yes, its probably useful to some and free, I am seeing its a bit of a stretch for the audience of this list.

```
The scope of this particular list is limited to things infrastructure developers (System Administrator, DevOps Practitioners, etc.) are likely to find useful. We love all the free services out there, but it would be good to keep it on topic. It's a bit of a grey line at times so this is a bit opinionated; do not be offended if I do not accept your contribution.
```

What do you think?

PR #1025, opened 2019-02-08
Added ApiFlash in the "APIs, Data and ML" section.

TimotheeJeannin:
ApiFlash has an unlimited free plan with 100 screenshots per month.


ripienaar:
thank you :)

PR #1027, opened 2019-02-15
Adding RequestBin.com

dylburger:
I noticed you had requestb.in listed here before. That service was shut down but requestbin.com is an equivalent tool with a few extra features.


ripienaar:
thank you

PR #1028, opened 2019-02-16
Add Shieldfy (shieldfy.io) in Security tools

netcode:
Adding Shieldfy in Security tools.

Shieldfy ( shieldfy.io )
Web application firewall and vulnerability detection for developers.

Thanks,


ripienaar:
thank you

PR #1031, opened 2019-02-19
Update README.md

okw95:
Added MailboxValidator, FraudLabs Pro and IP2Location as those provided both free and paid services.


ripienaar:
nice thanks

PR #1033, opened 2019-02-19
Added index.html for Github Pages

C0derLint:
## What's changed?
Added an index.html so that the repo could be served as a site using Github Pages.

### Why would we want to serve as a website?
 - Much easier to browse on mobile devices.
 - Integrated search means using Ctrl F is no longer needed
 - Cleaner UI. Site can also be later styled to look better.
 - Uses [docsify](https://github.com/docsifyjs/docsify), so the site is generated automatically from README.md
 - In the future, we could serve on a custom domain, but let's not get too ahead of ourselves :P
 - See a demo [here](https://c0derlint.github.io/free-for-dev).

What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,4 @@ rvm:
 before_script:
   - gem install awesome_bot
 script:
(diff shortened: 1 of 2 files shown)


ripienaar:
that looks sweet, thanks

PR #1046, opened 2019-03-01
'postgres.com' changed to 'progress.com'

mahmut-gundogdu:


ripienaar:
hah, nice one, completely missed that.  thanks!

PR #1052, opened 2019-03-18
Remove WeDeploy

BynariStar:
WeDeploy is being discontinued, more info can be found [here](https://wedeploy.com/blog/discontinuing-wedeploy/).


ripienaar:
thank you :)

PR #1054, opened 2019-03-31
adding reference to the 3 major cloud providers always-free services.…

kadamwolfe:
…  slight update to cloudflare and gitlab possibilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,8 @@ You can find the list on [GitHub](https://github.com/ripienaar/free-for-dev) or 
 Table of Contents
 =================
 
-   * [Major Cloud Providers Always-Free Limits](major-cloud-providers)
+   * [Major Cloud Providers Always-Free Limits](#major-cloud-providers)
+     * [Google Cloud Platform](#google-cloud-platform)
    * [Analytics, Events and Statistics](#analytics-events-and-statistics)
    * [APIs, Data and ML](#apis-data-and-ml)
    * [Artifact Repos](#artifact-repos)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -645,3 +648,8 @@ Table of Contents
   * [getawesomeness](https://getawesomeness.herokuapp.com) — Retrieve all amazing awesomeness from GitHub... a must see
   * [education.github.com](https://education.github.com/pack) — Collection of free services for students. Registration required
   * [Framacloud](https://degooglisons-internet.org/en/list/) — A list of Free/Libre Open Source Software and SaaS by the French non-profit [Framasoft](https://framasoft.org/en/).
+
+## Major Cloud Providers Always-Free limits
```
ripienaar: I think these can go in existing categories - and some of them are already in existing ones arent they?

README.md
```diff
@@ -619,6 +621,7 @@ Table of Contents
   * [canister.io](https://canister.io/) — 20 free private repositories for developers, 30 free private repositories for teams to build and store Docker images
   * [Whales](https://github.com/Gueils/whales) — A tool to automatically dockerize your applications for free.
   * [PWD](https://labs.play-with-docker.com/) — Play with Docker. A simple, interactive and fun playground to learn Docker
+  * [Gitlab](https://gitlab.com) - Per-repo container registry.  10GB limit.
```
ripienaar: prefer we list them just once

kadamwolfe: I get that, but they offer quite a lot of free services.  Which section would they all best fit under?

ripienaar: maybe a larger re-org for developer tooling is needed and we move a bunch of them in there?

README.md
kadamwolfe: Some of them can go in existing categories, yes.  Some of the items would probably need creation of their own as they are fairly unique (ex AWS Lambda or Azure AD).  They each provided a dozen or more disparate services for free so they would have many listings each.  

I'm not sure how these would be best organized.  If we want each item listed in a parent category that's fine, just takes time and would spin off quite a few new categories with a single item beneath.  Or we could just list out all services per provider under a separate provider heading?  

It gets a bit messy either way, which is why I opted to just say "hey, Google offers all of <link> for free." etc; just for the sake of having one listing and a reference to the official doc and the notion that they are subject to change.  Though this opens up a broader question as how to handle this for all free-tiers; like the Gitlab (VCS, CI/CD, issue tracking and more) discussion we are also having, or Cloudflare (DNS, CDN, Firewall etc) or anyone that offers multiple services that are in differing categories.  

I think, maybe, providers that offer more than one service deserve their own category with a "see <provider>" mention under each global category that applies.
I'm not set on that, but I think that's a decent way to go.

ripienaar: yeah i think having this category is right, we need to then pull these ones out of all the rest of the categories.

Maybe call it Cloud Native Infrastructure Providers and add it near the top of the list as they are major, pull all the similar ones to these in there and remove elsewhere.

Dont want to complicate things - cloudflare is iaas, just quite focussed/specific iaas around a theme.   When services drift into other areas we list them by their primary focus - gitlab example below, I'll just keep them under where they are now.  We have to be realistic about the limits of this format.

kadamwolfe: sounds good enough to me.  I'll work on this as I have time and send in an updated PR hopefully by the weekend.

ripienaar:
I'll merge as this is amazing as it is, if you had more tweaks in mind do send more PRs, thanks a lot @kadamwolfe

PR #1060, opened 2019-04-16
overv.io is no longe running as of Q2 2018

iandrewdunn:


lucca65:
-1, although I've used (and liked it) in the past, they are closing down https://blog.overv.io/attention-citizens-of-overvio-1d1957d253c

actually its been more then 1 year since they stopped developing it

ripienaar:
thank you!

PR #1062, opened 2019-04-27
Removed Down links.

aliglelo:
I have checked all the links and removed
https://sendtoinc.com/
as well as
https://inch-ci.org
because both these projects are down 
screenshot below:
sendtoinc
inch


ripienaar:
thank you

PR #1071, opened 2019-05-09
CI/CD section: update bitrise information and removed its duplicate

luanfsf:
Remove duplicated item in CI/CD section, also updates info about bitrise.


ripienaar:
thank you!

Issue #1073, opened 2019-05-13
webtask.io not-accepting new clients


lbonanomi:
<img width="1138" alt="webtask io no new clients" src="https://user-images.githubusercontent.com/5369016/57651041-484e1180-759a-11e9-8730-6de51332b920.png">

ripienaar:
thanks, want to do a PR to remove them?

PR #1074, opened 2019-05-14
remove webtask.io

lbonanomi:
Site webtask.io is not accepting new user signups.

This PR attempts to address https://github.com/ripienaar/free-for-dev/issues/1073


ripienaar:
Thank you :)

PR #1076, opened 2019-06-16
Add heliohost

imba-tjd:
Note: I didn't try it, only found it.
And I'm not sure whether it belongs to *Web Hosting*


ripienaar:
Thank you

PR #1084, opened 2019-07-06
Add remotemysql.com

bitsnaps:
Disclaimer: I do not belong to this service, I've just tested remotemysql and found it works as it says so far so good.


ripienaar:
nice one thank you

PR #1089, opened 2019-07-13
Removed cnverg.com.

aliglelo:
Removed https://www.cnverg.com/ because it is no longer available or down
as shown below:
www cnverg com_


ripienaar:
Thank you

PR #1093, opened 2019-07-18
Add gitea and codeberg

weyusi:
gitea.com and codeberg are hosted version of [Gitea](https://github.com/go-gitea/gitea), which is a community-fork of Gogs (used by notabug.org).


ripienaar:
Nice. I know gotta but somehow missed that they also do hosting. Thanks.

PR #1094, opened 2019-07-18
Move CI comparison to the top

weyusi:
and use user/repo as name. It has more complete list.

Displaying `github.com` as the name is a bit confusing, since it doesn't offer CI.


ripienaar:
Makes sense. Thanks!

PR #1101, opened 2019-08-02
Add Dnote to the "Miscellaneous" section

sungwoncho:
[Dnote](https://dnote.io) is an open source note-taking software offered as a hosted service. However, developers can use it for free by self-hosting.


ripienaar:
hmm, its a fine line but we dont accept systems where the free component is self hosting - editors and such we take - but this is a self hosted thing you have to put on a server, I don't think that will fit with this list, sorry.

Thanks a lot for the effort though

sungwoncho:
I see. Thanks for putting together and maintaining this list.

PR #1118, opened 2019-09-01
Changed appear.in to whereby.com

kreatemore:
Updated appear.in entry to use their new name: whereby.com


kreatemore:
Here's some background: https://medium.com/the-making-of-whereby/appear-in-is-now-whereby-b258d77d3ad9

> You will still be able to use appear.in room links up until September 1, 2019. After that point, appear.in links will no longer work and you’ll need to use your whereby.com link.

ripienaar:
👍 👍

PR #1119, opened 2019-09-03
Added zenserp.com to APIs, Data and ML

saasindustries:


ripienaar:
I'll pass on this one - scraping google search is against their TOS and robots.txt.

PR #1120, opened 2019-09-17
replace datamaps.co with datamaps.world

caspg:
Datamaps.world is a next version of the datamaps.co platform. It offers paid plans but it also includes a generous free tier.


ripienaar:
Thank you :)

PR #1128, opened 2019-09-23
Correct a typo in currencystack.io description

bxio:
Under the payments/billing section, currencystack was listed as 
`currencystack.io — Production-re3ady real-time exchange rates for 154 currencies.`

We remove the stray `3`.


ripienaar:
Thank you :)

PR #1129, opened 2019-09-24
Adding render.com

bitsnaps:
A unified platform to build and run all your apps and web app free SSL, a global CDN, private networks and auto deploys from Git, free for static web page


ripienaar:
nice, thank you

PR #1131, opened 2019-10-01
Added veriphone.io

BENADIL:
Added veriphone.io to International Mobile Number Verification API and SDK


ripienaar:
sweet thanks

PR #1132, opened 2019-10-01
Adding drone.io

franciscofsales:
Free self-hosted and super powerful container-based CI/CD system


ripienaar:
Drone is great but unfortunately this list is not for self hosted software. 

Thanks for taking the time to contribute :)

PR #1139, opened 2019-10-09
Remove duplicate line for JetBrains Youtrack

varesa:
Additionally update the information as the number of free users allowed has changed


ripienaar:
thank you

PR #1156, opened 2019-10-18
Add zenscrape.com

saasindustries:
Web scraping API with 1000 requests free/month.  Extra free credits for students and non-profits.


ripienaar:
thanks

PR #1159, opened 2019-10-18
Add stadiamaps.com as map data provider.

lseelenbinder:
Change and title says it all. Happy to answer any further questions!


ripienaar:
thank you

PR #1161, opened 2019-10-18
Adds Ably Realtime as a BaaS provider

remotekieran:
Ably serves the same developers as BaaS providers on the list like Pusher, PubNub, and Streamdata.io.


ripienaar:
looks awesome thanks

PR #1162, opened 2019-10-18
added transposit, serverless API app builder

mooreds:
I added Transposit.com, including details about the free tier.


ripienaar:
Hmm, this looks sweet however ```That's why apps built in the first six months of beta will be free to run on our platform.``` doesnt that mean its only 6 months free? Then it wont do for this list I am afraid

mooreds:
You are totally correct. I was under the impression that public apps would be free forever, but that doesn't appear be the case based on the pricing doc. My bad.

ripienaar:
no sweat, we appreciate the contribution regardless :)

PR #1170, opened 2019-10-18
Updated Contentful Information

Shy:
We've modified our free plan to make it include more features. Just updating the line about Contentful to reflect that.


ripienaar:
nice, thanks for keeping it current

PR #1174, opened 2019-10-18
Add link for browserstack open source pricing

JasonMFry:
Browserstack kinda hides this info under the fold and behind a couple of clicks, but having the direct link is much more convenient for users.


ripienaar:
good stuff thank you

PR #1181, opened 2019-10-19
Update README.md

halocameo:
Updating Bitbucket's description as they are removing support for Mercurial


ripienaar:
thanks for the update

PR #1183, opened 2019-10-19
Fix "Payment / Billing Integration" section link

mos3abof:
The link to the "Payment / Billing Integration" section in the table of contents
had an extra `-`. This commit fixes that.


ripienaar:
Thank you

PR #1184, opened 2019-10-19
Adding Clubhouse as an Issue Management resource

aniforprez:
Adding [Clubhouse](https://clubhouse.io/) as an addition to the Issue Tracking and Project Management resource. It is free for up to 10 users

Disclaimer: I do not in any capacity work for clubhouse


ripienaar:
thank you

PR #1185, opened 2019-10-19
Add Splunk to log management section

opw0011:
Splunk is a well known log management tool that worth mentioning.


ripienaar:
thank you

PR #1191, opened 2019-10-20
Added Cally.com, free scheduling made easy.

Dynom:


bitsnaps:
@ripienaar There should be a limit somehow, it must be mentioned somewhere.

PR #1192, opened 2019-10-20
Added Rapid API

Yusadolat:
Added Rapid API :zap:


ripienaar:
So using rapidapi is free but that doesnt imply anyitng about the APIs it connects you to? I think for this list I am more interested in the actual APIs than a broker.  Not sure, seems a bit of a stretch for the audience listed in the top of the README.  What do you think?

Yusadolat:
I understand your point, I have used rapidAPI for a recent project and i didn't pay it so I thought it might be of good use for other devs. I will close the PR

ripienaar:
thank you

PR #1194, opened 2019-10-20
Removed entryDNS

dimitarnestorov:
On the [home page](https://entrydns.net/):
> All users must pay a one-off registration fee of $10 (USD)


ripienaar:
thank you

PR #1195, opened 2019-10-20
Add devless BAAS Platform devless.io

koolamusic:
DevLess provides an easy to use Data Storage for web and mobile applications as well as application components so there's little need for developers to write custom code to handle complex functionalities of their app. Develop Less with DevLess.

https://devless.io


ripienaar:
SSL certificate expired ages ago? Also appears to be something you install in your own infra not a service someone offers.

PR #1197, opened 2019-10-20
Add pythonanywhere

philippG777:
Add pythonanywhere to hosting-section


bitsnaps:
pythonanywhere already exists in PaaS isn't that right section?

ripienaar:
Yeah, happy with where it is.  Thanks for the contribution though!

PR #1198, opened 2019-10-20
add openshift PaaS

sdil:


ripienaar:
thanks

bitsnaps:
are you sure it's not limited? https://github.com/ripienaar/free-for-dev/pull/1008

ripienaar:
yeah you're right I thought I remember it being on here at some point. I will revert the terms of their trial isnt clear at all

PR #1200, opened 2019-10-20
Added monitoring service

jovdra:
Regardless of the current beta testing phase, core functionality of the service will remain free for the foreseeable future, as stated on their site.


ripienaar:
thanks

PR #1208, opened 2019-10-21
Add "Fathom" to analytics

Pierstoval:
Been using it for a while now, and it's nice 😄 

> https://usefathom.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -702,6 +702,7 @@ Table of Contents
    * [Moesif](https://www.moesif.com) — API analytics for REST and GraphQL. (Free up to 500,000 API calls/mo)
    * [quantcast.com](https://www.quantcast.com/products/measure-audience-insights/) — Unlimited free analytics
    * [getinsights.io](https://getinsights.io) - Privacy-focused, cookie free analytics, free for up to 5k events/month.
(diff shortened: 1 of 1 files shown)


ripienaar:
looks nice, unfortunately this is a list specifically for software where the hosted SaaS version has a free tier.

Appreciate the contribution though!

maple3142:
But [README.md](https://github.com/ripienaar/free-for-dev/blob/fda5b684149b6fc9781aa8604793833394f92a6d/README.md#L13) said `This list is only for as-a-Service offerings, not for self-hosted software. For a service to be eligible it has to offer a Free Tier and not just a free trial. If the Free Tier is time bucketed it has to be at least a year.`. Perhaps this is not appropriate for this project?

@ripienaar Besides, I found that `opsdash.com` should be removed too. Because its `free for 5 servers and 5 services` is for [self-hosted](https://www.opsdash.com/pricing-self-hosted), not `as a Service`. Their hosted service only has [14-days trial](https://www.opsdash.com/pricing).

ripienaar:
@maple3142 thanks, it could be that they had a better free offering in the past - welcome to kill it via a PR since indeed it seems it does not belong anymore

Pierstoval:
Thanks for the clarification 🙂

PR #1212, opened 2019-10-21
fix: added Bearer.sh + ordered APIs tools by name

Frenchcooc:
Hey there,

I'm a developer advocate at [Bearer.sh](https://www.bearer.sh) which provides a generous free-tier for developers looking for an API integrations tool. The first 100K API requests are on us each month.

PS: I also took a bit of time to order the list of tools by name in that category.


ripienaar:
nice one, thank you

PR #1218, opened 2019-10-21
added ui resources

JoshuaPoddoku:


README.md
```diff
@@ -642,6 +642,8 @@ Table of Contents
   * [pexels.com](https://www.pexels.com/) - Free stock photos for commercial use. Has free API that allows you to search photos by keywords.
   * [whimsical.com](https://whimsical.com/) - Collaborative flowcharts, wireframes, sticky notes and mind maps. Create up to 4 free boards.
   * [lensdump.com](https://lensdump.com/) - Free cloud image hosting
+  * [freepik.com](https://freepik.com/) - Free stock and Premium Illustrations .Ai/ .eps /.psd
+  * [dribbble.com](https://dribbble.com/) - Inspiration for Designers
```
ripienaar: I do not think this one really fits with the purpose of the list, sure devs might want to host themselves there - but ultimately its just too far out from what this is after, freepik looks great though.

```The scope of this particular list is limited to things infrastructure developers (System Administrator, DevOps Practitioners, etc.) are likely to find useful. We love all the free services out there, but it would be good to keep it on topic. It's a bit of a grey line at times so this is a bit opinionated; do not be offended if I do not accept your contribution.```

Comment (JoshuaPoddoku) [reply README.md]: @ripienaar  Thank you for the feedback. I just wanted to help both developers and designers if they are on the same team.

File: README.md [hunk 2]
```diff
@@ -652,8 +652,12 @@ Table of Contents
   * [unsplash.com](https://unsplash.com/) - Free stock photos for commercial and noncommercial purposes (do-whatever-you-want license)
   * [whimsical.com](https://whimsical.com/) - Collaborative flowcharts, wireframes, sticky notes and mind maps. Create up to 4 free boards.
   * [lensdump.com](https://lensdump.com/) - Free cloud image hosting
+
```
ripienaar: do you mind removing these 3 empty lines? they will mess up the bullet list :)

PR #1219, opened 2019-10-21
add gitub pipelines coz it supports i.e github actions

iamsaalim:


ripienaar:
The thing about github actions is - we have no idea about the pricing they have not made to my knowledge any statement about the pricing, only that its free during beta.

Have you seen something to suggest there will be an ongoing free tier?

maple3142:
@ripienaar "Free during beta" only applies to private repos, so it is free for all public repos even after beta

ripienaar:
a cool, I see now - perhaps they updated the info since I looked previously, which was a while ago.

Thank you!

PR #1231, opened 2019-10-22
Remove landscape.io service, it's down for couple months

zgoda:
It shows "Back soon" message and nothing else.


ripienaar:
thank you

PR #1232, opened 2019-10-22
Add ZEIT Now to Web Hosting

msweeneydev:
This PR adds ZEIT Now to the Web Hosting section.


ripienaar:
nice one, thanks

maple3142:
Is it okay that this entry exists in both Web Hosting and PaaS section?

ripienaar:
nope, I somehow missed it was there already thanks for the headsup

PR #1233, opened 2019-10-22
Add Adobe XD and Unsplash

Zahnstochermann:
Add Unsplash, which is a very powerful free stock photo library that even has an API. 
Add Adobe XD which is a Wireframe & Prototyping tool.


ripienaar:
thank you

PR #1237, opened 2019-10-23
Add CodersRank - The Ultimate Profile For Coders

ilyakaznacheev:
[CodersRank](https://codersrank.io/) - is a free platform for developers, which measures your skills and knowledge based on different sources like GitHub, GitLab, StackOverflow, and other sources and provides a nice-looking profile with deep knowledge analytics and skill comparison.


ripienaar:
appreciate the effort to add it, unfortunately I do not think it fits within the context of this list

PR #1242, opened 2019-10-23
Add Todoist and Twist

aladagemre:
Todoist is the top productivity app for task management. Twist is the async team communication app from the same company.


ripienaar:
thanks

PR #1245, opened 2019-10-25
Add GitHub Actions to CI / CD

skoshy:
Been using GitHub actions for CI/CD and it has been great, I think it should be added to the list!


ripienaar:
The main github entry already has it `Unlimited public repositories and unlimited private repositories (up to 3 collaborators) with Pipelines for CI/CD`, I dont really want to add same thing many places

thanks for the contribution!

PR #1253, opened 2019-10-28
Update Transloadit offer

kvz:
Hi, I'm the founder of Transloadit and since we've been added, we've expanded our free offering. Just leaving this update to reflect our latest


ripienaar:
very nice, thank you

PR #1257, opened 2019-10-29
add Mailnesia disposable email

rizdaprasetya:
Added another free disposable temporary mail service, with unique feature of auto visiting registration link.


ripienaar:
thanks

PR #1264, opened 2019-10-30
fix broken link without http protocol

rizdaprasetya:
Few links are without `https://` protocol, which result in redirect to relative path (broken link). So I added `https://` protocol.


ripienaar:
thanks

PR #1283, opened 2019-11-01
Add FusionAuth to Security and PKI section

trex:
Adding FusionAuth as a CIAM, SSO, user management solution to the "Security and PKI" section


README.md
```diff
@@ -301,6 +301,7 @@ Table of Contents
   * [snyk.io](https://snyk.io) — Can find and fix known security vulnerabilities in your open source dependencies. Unlimited tests and remediation for open source projects. Limited to 200 tests/month for your private projects.   
   * [letsencrypt.org](https://letsencrypt.org/) — Free SSL Certificate Authority with certs trusted by all major browsers
   * [globalsign.com](https://www.globalsign.com/en/ssl/ssl-open-source/) — Free SSL certificates for Open Source
+  * [FusionAuth](https://fusionauth.io/) - Built for devs: Integrates with every platform, every codebase. Complete authentication, registration, login, MFA, SSO and more. Download for free.
```
ripienaar: this looks sweet but I think the free option here is only self hosted? This list is only for software where the free component is in the SaaS

trex: You are correct, free for self-hosted, and not for SaaS.

ripienaar: thanks, really good looking but unfortunately not for this list.  Thanks for contributing!

PR #1288, opened 2019-11-03
Add Tefter.io under team collaboration

zorbash:
Tefter is a social bookmarking app which enables team collaboration through Slack. It's free, without any limitations for open-source teams.


ripienaar:
thanks

PR #1291, opened 2019-11-04
Add portainer - Making Docker management easy

jnadaud:


README.md
```diff
@@ -780,6 +780,7 @@ Table of Contents
   * [Whales](https://github.com/Gueils/whales) — A tool to automatically dockerize your applications for free.
   * [PWD](https://labs.play-with-docker.com/) — Play with Docker. A simple, interactive and fun playground to learn Docker
   * [Gitlab](https://gitlab.com) - Per-repo container registry.  10GB limit.
+  * [portainer.io](https://www.portainer.io) - Build and manage your Docker environments with ease today
```
ripienaar: this appears to be self hosted software right? You install both the agent and the server into your own swarm. 

This list is for software as a service that you do not self host, so I dont think it would work?

jnadaud: Ok, noted!

ripienaar:
thanks for the effort though, really appreciate it :)

PR #1293, opened 2019-11-08
Remove Librato.com

gmasse:
Librato is now part of SolarWinds. They appear to no longer provide free tier, only free trial.


ripienaar:
Thank you

PR #1298, opened 2019-11-11
Add QuickChart

typpo:
Added the QuickChart web service free tier.  It also happens to be open source (https://github.com/typpo/quickchart)

Thanks :)


ripienaar:
thanks

PR #1301, opened 2019-11-12
Add dbdesigner.net

bitsnaps:
dbdesigner.net is free cloud based Database schema design and modeling tool with a free starter.


ripienaar:
thank you

PR #1302, opened 2019-11-12
Swiftype no longer offer a free plan

Accudio:
Source: https://brownbox.net.au/swiftype-silently-drops-free-and-low-cost-plans/


ripienaar:
thank you

PR #1303, opened 2019-11-13
Doodle does not offer Free Tier anymore

altinburak:
Doodle stops offering Free Tier anymore, it has only free trial.


ripienaar:
thank you

PR #1305, opened 2019-11-15
Deleted ProFreeHost

iondrey:
ProFreeHost uses a free reselling program from Byet no different from it


ripienaar:
Number of low fairly dodgy looking free hosts I think I shouldn’t have accepted in retrospect. 

Thanks for the edit

PR #1306, opened 2019-11-15
Removed Byet arbitrary hosting

iondrey:
InfinityFree and 2HOST use Byet’s free reselling program, no different from them

Sorry for the two commits, did not immediately notice the rest


ripienaar:
Awesome. Thanks.

PR #1307, opened 2019-11-17
add kubesail service and fix a typo

sdil:


ripienaar:
Weird. Thought we had this already. Thanks!

PR #1308, opened 2019-11-18
Add configcat.com to Miscellaneous section. This is a feature flag service supporting kill switches, gradual rollouts and trunk based development.

david-zoltan:


ripienaar:
Nine. Thanks.

PR #1318, opened 2019-11-25
Add Scaleway (75 GB Free S3-Compatible Storage)

brentybh:
S3-Compatible Object Storage. Free 75 GB storage and external outgoing traffic


ripienaar:
very nice

PR #1320, opened 2019-11-25
Added a JSON storage service

cyrilbois:
Added extendsclass.com, a free JSON storage service, to "Storage and Media Processing"


ripienaar:
this looks nice and all but its too vague about what you actually get, its not something you can build on.  The intention of this list is that devs who want to develop stuff using services dont need to pay to do the development - later they can offer clients a solid service by then only paying for it. Well thats the rough charter, its a bit of a grey line.

This service has no paid offering, no clear limitation, sketchy/no authentication etc, its a nice toy but not something to build on I think?

cyrilbois:
Yes you are right!

ripienaar:
thanks, unfortunately not a fit for the list then, appreciate the effort though!

PR #1322, opened 2019-11-26
added some design tools and periods

SaijoGeorge:
added codemyui , allthefreestock to the design section and a bunch of periods for entries missing them.


ripienaar:
thanks

PR #1325, opened 2019-11-26
Added AnonAddy to Email

willbrowningme:
AnonAddy has a free plan that allows you to create unlimited email aliases, it is also open-source and self hostable.


ripienaar:
thanks

PR #1326, opened 2019-11-26
Adding Postput: Free microservice for file storage and processing

deviarchscs:


ripienaar:
looks nice, unfortunately this list is for Services delivered in a aaS model, not self hosted software, this appears to not have any hosted option to me?

PR #1343, opened 2019-12-02
chore: added myfakeapi to API services

angad777:
Added [myfakeapi](https://myfakeapi.com) under free to use API resources.


ripienaar:
nice thanks

PR #1345, opened 2019-12-04
Add TailWind to Design and UI section

zOmegad:


README.md
```diff
@@ -716,6 +716,7 @@ Table of Contents
   * [Adobe XD](https://www.adobe.com/products/xd.html) - Wireframe & Prototyping tool similar to Sketch. Free plan covers: 1 active shared design spec, Adobe Fonts Free (limited set of fonts), 2GB of cloud storage.
   * [AllTheFreeStock](https://allthefreestock.com) - a curated list of free stock images, audio and videos.
   * [CodeMyUI](https://codemyui.com) - Handpicked collection of Web Design & UI Inspiration with Code Snippets.
+  * [TailWind CSS](https://tailwindcss.com/) - A utility-first CSS framework for rapidly building custom designs.
```
ripienaar: nice but I am afraid I dont think it fits with this list, its not a service really, and I think perhaps one or two others in the section should also be killed for the same reason but thats for another time

thanks for the effort though

PR #1346, opened 2019-12-05
Add Filestash

mickael-kerjean:
Hi,

I wrote that software which I think is awesome. I hope it is worthy of your selection!

- [X] This is a list of software (SaaS, PaaS, IaaS, etc.) and other offerings that have free tiers for developers. `The free tier is the online demo which can be found from the homepage of the website: https://demo.filestash.app There's an upload limit set to 10MB on it`
- [X] things infrastructure developers (System Administrator, DevOps Practitioners, etc.) are likely to find useful `I don't know a sys admin or devops who never used any of S3, FTP, SFTP, GIT, WebDAV and more. Also I've noticed that the nice part with Filestash when it comes to devops is collaborating with people who don't have much technical knowledge as those people need tool that's easy to use (as Dropbox)`
- [ ] This list is only for as-a-Service offerings, not for self-hosted software. `We're doing both self-hosted and as-a-Service. The as-a-Service version is simply hosted on our servers. I hope this is ok?`

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ Table of Contents
 
 ## Tools for Teams and Collaboration
 
+  * [Filestash](https://www.filestash.app) — A Dropbox-like file manager that connects to a range of protocols and platforms: S3, FTP, SFTP, Minio, Git, WebDAV, Backblaze, LDAP and more.
   * [GitDuck](https://gitduck.com/) — Private live coding and remote pair programming for distributed teams. Open-source tool
   * [scinote.net](https://scinote.net) — Scientific data management and team collaboration. One Team with unlimited number of users, backup and 1 GB storage space
   * [whereby.com](https://whereby.com/) — One click video conversations, for free (formerly known as appear.in)


ripienaar:
looks nice, however the free cloud instance is just a shared demo that everyone use right? it's not private?

I dont think that will pass the bar for a SaaS tool with a free offering I am afraid

mickael-kerjean:
> the free cloud instance is just a shared demo that everyone use right?

Yes. Also the demo is the full product except users don't have access to the admin console and a few other limitations that are all details on the pricing page: https://www.filestash.app/pricing/

It is quite the same idea as a few others tools that are publish here such as taiga and gitlab where the free offering is quite literally an instance of an open source product that's shared by everyone + a few additional arbitrary limits that are lifted up when becoming a customer.

> it's not private?

It is assume the private part is the backend you're connecting to, be it an FTP server or a S3 bucket. This is central to why the project exists in the first place and is the major difference with say seafile that's shown here.

Also, compared to some other products shown here such as draw.io and jitsi, there's quite a lot of things that's handle on the backend side as well: session management, database of shared link, cache management, ...)

> I dont think that will pass the bar for a SaaS tool with a free offering I am afraid

As shown on the pricing page (https://www.filestash.app/pricing/), the demo instance is beeing actively marketed as our free saas

ripienaar:
thanks

PR #1349, opened 2019-12-08
Fixed minor typo that broke Postwoman link

bdnh:
- A space was placed between the `[ ]` and `( )` of the Postwoman markdown link, causing it to appear broken on the front end


ripienaar:
Thank you!

PR #1352, opened 2019-12-13
Added HashiCorp Vault

Necromancerx:
https://www.vaultproject.io/
Manage Secrets and Protect Sensitive Data 
Secure, store and tightly control access to tokens, passwords, certificates, encryption keys for protecting secrets and other sensitive data using a UI, CLI, or HTTP API.


ripienaar:
while vault is great, this is a list for SaaS offerings, so for example if there is a free tier in the hashicorp cloud product that could be included here but not vault on its own as its self hosted software

PR #1355, opened 2019-12-16
fixes incorrect myfakeapi.com indention

jontonsoup4:
_myfakeapi.com_ was showing up as a nested list rather than a list item because of an extra space before the `*`. 

Screenshot of original problem:
Screen Shot 2019-12-16 at 2 05 57 AM


ripienaar:
thank you

PR #1374, opened 2019-12-28
Add Emacs to the list

lokedhs:


ripienaar:
yeah sure, its a downloadable editor but I dont think I'm going to add it here sorry, dont have a strong reason as such - the editors here are already outside of the charter of the list and adding core unix utils is way past what this list is for.

PR #1377, opened 2019-12-29
Updated okteto details

TxSadhu:
Okteto now provides 3 namespace instead of 5, and added some deatils that were not there previously like 150 Disk space


ripienaar:
Thank you

PR #1381, opened 2020-01-02
Add InfluxDB service

sdil:


ripienaar:
also seems to be a bit out of date, if you can pls make against latest master that'd be great - you can do both in the same PR if you want

sdil:
> also seems to be a bit out of date, if you can pls make against latest master that'd be great - you can do both in the same PR if you want

Fixed. Updated to latest code :)

PR #1390, opened 2020-01-07
Kuber.host have no real free plan anymore :(

Nek-:
I know there's still a free plan you can have but it shutdown automatically after 30 days. It's more testing plan than free plan.

So I think it should be removed from the list. Sad.


ripienaar:
thank you

PR #1395, opened 2020-01-08
Add Prismic Headless CMS

frvncis:


ripienaar:
thank you

frvncis:
Thank you for that list.
I feel that we could add a "CMS" category to the list since I saw that
there are quite a few CMSes in the list and it would make it clearer. Would
that be OK if I do it in another PR?
Thanks!
*François*
<http://www.bnl.services>


Le mer. 8 janv. 2020 à 16:22, R.I.Pienaar <notifications@github.com> a
écrit :

ripienaar:
yup, feel free to pull those out! thanks

PR #1400, opened 2020-01-14
Added letsmonitor.org to monitoring services

arulprabakaran:
Added letsmonitor.org to monitoring services


ripienaar:
thank you

PR #1402, opened 2020-01-15
Added codenotary.io to Security and PKI

dzlabsch:
Added codenotary.io to Security and PKI

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,6 +327,7 @@ Table of Contents
   * [auth0.com](https://auth0.com/) — Hosted free for development SSO. Up to 2 social identity providers for closed-source projects.
   * [bitninja.io](https://bitninja.io/) — Botnet protection through a blacklist, free plan only reports limited information on each attack
   * [cloudsploit.com](https://cloudsploit.com/) — Amazon Web Services (AWS) security and compliance auditing and monitoring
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -327,6 +327,7 @@ Table of Contents
   * [auth0.com](https://auth0.com/) — Hosted free for development SSO. Up to 2 social identity providers for closed-source projects.
   * [bitninja.io](https://bitninja.io/) — Botnet protection through a blacklist, free plan only reports limited information on each attack
   * [cloudsploit.com](https://cloudsploit.com/) — Amazon Web Services (AWS) security and compliance auditing and monitoring
+  * [CodeNotary.io](https://www.codenotary.io/) — Open Source platform with indelible proof to notarize code, files, directories or container
```
ripienaar: while indeed it says "start for free" I cannot quickly find what that actually gives me, is there a url with the info about the free features?

dzlabsch: the service on CodeNotary.io is completely free (therefore no pricing) and you can notarize any kind of digital asset using the CLI or API. The only limitation is fair use, so no simultaneous notarization of many assets the same second.

ripienaar: Thanks

PR #1408, opened 2020-01-21
Added stackbit.com

Snickdx:
https://www.stackbit.com/about/


ripienaar:
There's no pricing information and only what appears to be a early access preview type thing? That's not suitable for this list.

Snickdx:
The trial (not early access) seems to be free no credit card required. I've been using it for a while but I admit I'm not sure if It will be like that forever.

ripienaar:
Yeah without a clear future indication it's hard, README says:

```
NOTE: This list is only for as-a-Service offerings, not for self-hosted software. For a service to be eligible it has to offer a Free Tier and not just a free trial. If the Free Tier is time bucketed it has to be at least a year.
```

else they change their plan and it gets orphaned here

Snickdx:
Understood will make another PR if they publish a free trier

PR #1413, opened 2020-01-23
Remove SciNote

blalond:
SciNote was under tools-for-teams-and-collaboration but free level is now just for 1 user.


ripienaar:
thank you

PR #1422, opened 2020-02-10
Documentation

wallacemyem:
Add https://docs.brickng.com is a startup for free knowledge base and documentation hosting based on Drupal Vardoc. I recommend it under the miscellaneous tab.


ripienaar:
Apart from this PR being against some out of date version of the docs I am just not seeing how this service works, there doesn't appear to be much by way of examples of what it does and the few articles linked from the top landing page - they just seem not like project documentation?

Do you have an example of what this does? And then perhaps update the PR against latest master

PR #1436, opened 2020-02-23
Keybase and Bitwarden services added

Ampsicora:


README.md
```diff
@@ -131,434 +131,436 @@ Table of Contents
 
 ## Source Code Repos
 
-  * [Azure DevOps](https://azure.microsoft.com/services/devops/) — Unlimited private repos (Git and TFS) for up to 5 users/team
```
ripienaar: why remove all these?

iondrey: I think this is a desire to make a single format

ripienaar: Judging by the size of the commit it was already pretty uniform :P and adding the / means fewer redirects when visiting the sites.

Ampsicora: > why remove all these?

Removing the "/"  was just a proposal to better uniform links, feel free to ignore it if you prefer

ripienaar: You'd need to remove it from the PR, I cant selectively ignore that

PR #1444, opened 2020-02-26
A suggestion to sort the list using this script

bitsnaps:
This is file is just a suggestion, so if a GoLang developer can makes it works for this repo, I think this gonna be a great way to sort the list by the number of stars of each repo.


ripienaar:
Stars per repo? None of the things we are linking to are things with stars.

PR #1445, opened 2020-03-03
Add GitHub Actions to CI/CD section

peckjon:


ripienaar:
Thanks for the contribution, unfortunately it's already listed the main github entry and we try not to duplicate things

peckjon:
Hrm, might I suggest a cross-reference system then (e.g. links from each subsection back to those main entries where appropriate)?  Readers tend to link down directly to the section they're interested in, meaning that they'll miss out on knowing about options unless they carefully read all entries in every single section.

peckjon:
Also confused @ripienaar as to why you'd apply this logic to GitHub but not GitLab and Bitbucket (which are also duplicated)?

ripienaar:
this might be the case but it would be basically the same as a duplicate entry - for it to have value the reference would need to have the same info.  I get it, it's not perfect - this is after all just a long list of links.  But it is what it is and I don't want to make it just much longer by adding dupes

peckjon:
Also confused @ripienaar as to why you'd apply this logic to GitHub but not GitLab and Bitbucket (which are also duplicated)?

ripienaar:
github too is listed multiple times already, if anything we should dedupe it.

peckjon:
Agreed, would like to see consistency one way or the other. That's not a PR I can submit, tho; would give the appearance of bias (I'm a GitHub employee).

ripienaar:
I wouldnt think you're biased :) but sure, I get it.

PR #1446, opened 2020-03-04
New "Font" section with 3 links

SaulSilver:
Personally as a web developer, I continuously need font resources such as the ones I have listed in the PR.


ripienaar:
cool, a bit on the edge if this fits the purpose of the list but its handy so will take it :)

SaulSilver:
Yeah felt so too. Thanks for the merge though 👍

PR #1450, opened 2020-03-06
Updated link description

rodenacker:
Sorry for the hassel - my md is not very good. WOuld like to update the description for my link


ripienaar:
no hassle :)

php-coder:
@rodenacker by the way, I'm using https://dillinger.io in such cases and found it useful. JFYI.

rodenacker:
Thanks, Slava

PR #1452, opened 2020-03-09
Add Howuku into Analytics and Visitor Recordings section

donaldng:
I would like to add my new analytics platform [howuku.com](https://howuku.com) into the list which comes with a freemium plan.

### DIFF
--- a/README.md
+++ b/README.md
@@ -871,7 +871,6 @@ Table of Contents
 
 ## Visitor Session Recording
 
-   * [howuku.com](https://howuku.com) — 200 sessions/month with 1 month data retention.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -870,6 +871,7 @@ Table of Contents
 
 ## Visitor Session Recording
 
+   * [howuku.com](https://howuku.com) — 200 sessions/month with 1 month data retention.
```
ripienaar: I appreciate your tool does a lot and do belong to both sections, but we tend to list only major things like github in multiple sections, so please pick one - else the list gets emmense

donaldng: Thanks for responding to my request. As per request, I've remove howuku listing from Visitor Session Recording section.

ripienaar:
thank you

PR #1458, opened 2020-03-14
Adding LogRocket to crash and exception management

callmegar:
Adding LogRocket, a nice crash and session replay tool for javascript, free for 1,000 sessions a month and integrates very well with sentry.io


ripienaar:
thanks

PR #1464, opened 2020-03-23
Adds Platform9 on PaaS section

Ziggoto:
I just found this new K8S provider by reading this [arcticle](https://www.infoq.com/news/2020/03/platform9-free-SAAS-kubernetes/)


ripienaar:
thank you

imba-tjd:
Firstly I havn't register this but I think it's not a free k8s cluster. A free plain comes with 20 nodes? That's too much. As waht I understand, it's only a control panel.

Evidence:

* The picture in https://platform9.com/sandbox/kubernetes/ asks you to choose infrastructure
* In https://platform9.com/blog/announcing-new-platform9-managed-kubernetes-pmk-plans-that-start-at-zero-cost-and-scale-as-customers-grow/ , they said "Deploy clusters in any environment ranging from developer laptops, on-premises VM’s or bare metal servers to edge infrastructure or public clouds"

Again, I havn't register this.

ripienaar:
that was my understanding too, perhaps we can improve the description to call this out better.

As a service managing k8s is still quite a valid service for this list so I am ok with the features

PR #1466, opened 2020-03-29
ShareX - Screen capture, file sharing and productivity tool

danilodene:
ShareX - Screen capture, file sharing and productivity tool 

https://getsharex.com/


ripienaar:
Thanks for the PR but ultimately I don't think this really works for this list, its desktop software that integrates with many other systems, but on its own its not a service

thanks!

PR #1467, opened 2020-04-04
Add LeanCloud

brentybh:
Mobile backend. 1GB of data storage, 256MB instance, 3K API requests/day, 10K pushes/day are free. (API is very similar to Parse Platform)

https://leancloud.app/pricing/
https://docs.leancloud.app/leanengine_plan.html#hash-849507193

Their DC is located in Los Angeles.
I am using the free plan on their China site since 2015, and works fine now.


ripienaar:
thank you

PR #1470, opened 2020-04-06
Added ilograph diagrams and Kumu graphs

Klendat:
I've been testing these lately. They're pretty neat and useful.


ripienaar:
thank you

PR #1471, opened 2020-04-06
Update README.md

hessam71d:
firebase hosting added

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,6 +616,7 @@ Table of Contents
 
 ## Web Hosting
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -616,6 +616,7 @@ Table of Contents
 
 ## Web Hosting
 
+  * [firebase](https://firebase.google.com/products/hosting) — Static, 1 GB free storage with 10 GB/month Bandwidth.
```
ripienaar, line 619: I think this would go with the section about google cloud at the top - plus its already listed in PaaS.

ripienaar:
If you would like to update this to move the one from PaaS into the google cloud section that would be great, else feel free to close.

thanks for your contribution!

PR #1473, opened 2020-04-07
Fix a biased description of honeybadger.io

Wuzado:
In my opinion, lists like this should be as neutral as they can be. Ads should generally be avoided in a case like that.


ripienaar:
thank you

PR #1476, opened 2020-04-09
Added Namecheap Supersonic

iondrey:


ripienaar:
nice thanks

brentybh:
@ripienaar This service only allows domain registered with NameCheap...

iondrey:
> @ripienaar This service only allows domain registered with NameCheap...

Third-party domains are supported only when using [FreeDNS](https://www.namecheap.com/domains/freedns/).

brentybh:
> > @ripienaar This service only allows domain registered with NameCheap...
> 
> Third-party domains are supported only when using [FreeDNS](https://www.namecheap.com/domains/freedns/).

OK, Cloudflare also requires changing the nameservers...
So I can't use it even if my domain is registered with NC but is using another DNS provider?

(The CDN is actually StackPath, which is using Anycast IP, just like Cloudflare. This means there's no CNAME record to add)

iondrey:
> > > @ripienaar This service only allows domain registered with NameCheap...
> > 
> > 
> > Third-party domains are supported only when using [FreeDNS](https://www.namecheap.com/domains/freedns/).
> 
> OK, Cloudflare also requires changing the nameservers...
> So I can't use it even if my domain is registered with NC but is using another DNS provider?
> 
> (The CDN is actually StackPath, which is using Anycast IP, just like Cloudflare. This means there's no CNAME record to add)

As practice has shown, it is not necessary to transfer the domain to their NS records; the main thing is to add the domain to your account. Then, in the Supersonic panel, apply the setting of DNS records and copy the established records to your DNS hosting.

brentybh:
Its features for free plan are awful but it worked.
Thanks, @iondrey !

PR #1480, opened 2020-04-19
Github - private repositories with unlimited collaborators now

JerryGoyal:
Github recently announced announce that they're making private repositories with unlimited collaborators available to all GitHub accounts.
Reference: https://github.blog/2020-04-14-github-is-now-free-for-teams/


ripienaar:
thank you

PR #1483, opened 2020-04-20
Zeit/now.sh free tier update

brentybh:
https://zeit.co/blog/simpler-pricing

https://zeit.co/docs/v2/platform/fair-use-policy
There guideline now says "Up to 100GB"


ripienaar:
thanks

PR #1488, opened 2020-04-23
Remove Selectel DNS Geo zone

brentybh:
Seems that they don't provide GeoDNS...
https://kb.selectel.com/docs/cloud-services/dns-hosting/dns_hosting/


ripienaar:
thanks

PR #1493, opened 2020-04-25
Removed formaholic

danielro:
I check the Forms section regularly and I've seen this message on formaholic for some time (couple of months) now:

"Signups are currently disabled due to abuse"


ripienaar:
thank you

PR #1495, opened 2020-04-27
Remove rollbar.

VictorioBerra:
Rollbar is no longer free for dev, its only a 15 day trial now.

I just tested this myself a couple weeks ago and got this email from them today:

image

The button says "trial" in it. 

image

When you click the button, the next modal clearly says "Start your free 14-day trial of Rollbar":

image


ripienaar:
thanks for the update!

PR #1496, opened 2020-04-27
Move Scaledrone from WebRTC to Messaging

herkyl:
Hey!

Proposing moving Scaledrone under the messaging subsection where it fits better.


ripienaar:
thanks for the update

PR #1515, opened 2020-05-19
Deleted Fiwit.io Entry

koehlertimo:
Deleted Fiwit.io because they shutdown their service!


ripienaar:
thank you

koehlertimo:
no problem!

PR #1519, opened 2020-05-20
Added Firebase Authentication to Security and PKI

DoodlesEpic:
I think it's a great addition to the list, unlimited free authentication, and users, it also provides up to 10 thousand phone verifications per month, very simple to implement, and certainly a great way to manage user authentication.

There is another description from the website which could be used, instead of the one I adapted: "Firebase Auth offers multiple methods to authenticate, including email and password, third-party providers like Google or Facebook, and using your existing account system directly. Build your own interface, or take advantage of our open-source, fully customizable UI."


ripienaar:
thanks

PR #1520, opened 2020-05-20
Update README.md

manthankool:
Added another FREE API for scraping.


ripienaar:
I reverted this PR, I dont think this is a free tier service.

manthankool:
This is a free service upto 1000 free credits.

ripienaar:
I see `Get started with 1000 calls for FREE per month` can you add that to the description initially I only saw 1000 credits, not that it's ongoing and every month.

manthankool:
O have added "free pack available"

PR #1526, opened 2020-05-28
Update Readme.md

Huseyinnurbaki:
Lifetime free mock server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ Table of Contents
   * [ip-api](https://ip-api.com) — IP Geolocation API, Free for non-commercial use, no API key required, limited to 45 req/minute from the same IP address for the free plan.
   * [APIFull](https://apifull.com/) — Web Scraping API handling headless browsers and rotating proxies for developers.
   * [WebScraping.AI](https://webscraping.ai) - Simple Web Scraping API with built-in parsing, Chrome rendering and proxies. 5000 free API calls per month.
+  * [Mocktail](https://hub.docker.com/r/hhaluk/mocktail) - Mocktail is a completely free, lightweight, ready to run containerized mock server with nice interface. No request limitations or paid subscriptions.
 
 ## Artifact Repos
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,6 +207,7 @@ Table of Contents
   * [ip-api](https://ip-api.com) — IP Geolocation API, Free for non-commercial use, no API key required, limited to 45 req/minute from the same IP address for the free plan.
   * [APIFull](https://apifull.com/) — Web Scraping API handling headless browsers and rotating proxies for developers.
   * [WebScraping.AI](https://webscraping.ai) - Simple Web Scraping API with built-in parsing, Chrome rendering and proxies. 5000 free API calls per month.
+  * [Mocktail](https://hub.docker.com/r/hhaluk/mocktail) - Mocktail is a completely free, lightweight, ready to run containerized mock server with nice interface. No request limitations or paid subscriptions.
```
ripienaar, line 210: So whats the SaaS aspect to this?

Huseyinnurbaki: Well, it may not be considered as SaaS. But the aim is it is actually free unlike the most other mocking tools. You can deploy it anywhere you want with a single command. The reason behind why I didn't deploy it and let people access, sign up and benefit from it is,  there is no point on that, mock servers are not usually needed for a long time, and you can run this one either on cloud or local machine. Plus why would I want to pay for mocking requests ? This saves time and money. Run it, use it, get rid of it. And then if you need it again start over, share apis etc.

ripienaar: I appreciate what you're saying, but if I list every free app on the plant I would have 10s of thousands of entries here.  So the list has to focus, and what it focus on is SaaS with a few exceptions like IDEs

Huseyinnurbaki: I see, you are right, I'll close it then. Thanks for the response.
But If you consider making a list of free applications, please let me know 😃

PR #1550, opened 2020-06-28
Remove jsonstore (service has shut down)

shu8:
This PR removes [jsonstore.io](https://www.jsonstore.io/) from the "Storage and Media Processing" section, as it's shut down:

image


ripienaar:
thank you

PR #1553, opened 2020-07-01
Add Code Time

brettmstevens7:


README.md
```diff
@@ -878,6 +878,7 @@ Table of Contents
    * [cocalc.com](https://cocalc.com/) — (formerly SageMathCloud at cloud.sagemath.com) — Collaborative calculation in the cloud. Browser access to full Ubuntu with built-in collaboration and lots of free software for mathematics, science, data science, preinstalled: Python, LaTeX, Jupyter Notebooks, SageMath, scikitlearn, etc.
    * [codepen.io](https://codepen.io/) — CodePen is a playground for the front end side of the web.
    * [codesandbox.io](https://codesandbox.io/) — Online Playground for React, Vue, Angular, Preact and more.
+   * [Code Time](https://www.software.com/code-time) - an extension for time-tracking and coding metrics in VS Code, Atom, IntelliJ, Sublime Text, and more.
```
ripienaar, line 881: hmm, bit of a blurry line but to me I dont think this suits the list. Yeah we list IDEs which are obviously dev only tools but mainly this is about SaaS for building customer services, this one is a bit of a odd one out here

sorry, thanks for the contribution though

brettmstevens7: Ok, thanks. I see what you mean, but there are plenty of tools in this list that aren't direct elements of a tech stack. Seems like a difficult line to draw. Could I add this to Miscellaneous or a new section, Productivity Tools?

ripienaar:
Hmm yeah put in misc, only new sections if we can put a few things in there else a new section isn’t solving a problem :)

ripienaar:
Hmm yeah put in misc, only new sections if we can put a few things in there else a new section isn’t solving a problem :)

PR #1564, opened 2020-07-06
Add staroid.com to the list of PaaS

Leemoonsoo:
I'd like to add staroid.com in the PaaS section.
https://docs.staroid.com/project/testdrive.html


ripienaar:
thank you

PR #1569, opened 2020-07-08
Update Kontent plan specification

Simply007:
The starter plan was transformed into the Developer plan to fits the needs. 
The biggest change for developers is the availability of Management API and Custom elements.

👍 for the website and list Kentico Kontent @ripienaar


ripienaar:
thank you

PR #1573, opened 2020-07-09
Add weblate.org

bitsnaps:
Web-based continuous localization, Copylefted libre software, used by over 1150 libre software projects and companies in over 115 countries.
P.S. it seems to be widely used, but not tested yet.


ripienaar:
thanks

PR #1580, opened 2020-07-17
Added Jelastic PaaS

SlavaKatiukha:


README.md
```diff
@@ -601,6 +601,7 @@ Table of Contents
   * [gigalixir.com](https://gigalixir.com/) - Gigalixir provide 1 free instance that never sleeps, and free-tier PostgreSQL database limited to 2 connections, 10, 000 rows and no backups, for Elixir/Phoenix apps.
   * [glitch.com](https://glitch.com/) — Free unlimited public/private hosting with features such as code sharing and real-time collaboration
   * [heroku.com](https://www.heroku.com/) — Host your apps in the cloud, free for single process apps
+  * [Jelastic](https://jelastic.com/) — Multi-Cloud Platform-as-a-Service that automates environment provisioning, scaling, clustering and security updates of traditional and cloud-native applications running inside containers. The platform is provided with free trial (usually for 14 days) across a network of local service providers, as well as can be installed for free from DigitalOcean marketplace running up to 10 nodes (Max 64 GB RAM each)
```
ripienaar, line 604: As per the top of the readme, short free trials do not qualify for this list.  Do they have a free tier?

SlavaKatiukha:
```suggestion
  * [Jelastic](https://jelastic.com/) — Multi-Cloud Platform-as-a-Service that automates environment provisioning, scaling, clustering and security updates of traditional and cloud-native applications running inside containers. The platform is provided across a network of local service providers with a set of free resources per hour (RAM&CPU, disk, network, public IP), the amount of free resources depends on the provider and can be checked in the [cloud catalog](https://jelastic.cloud/?freeResources=CLOUDLETS_STORAGE_NETWORK_IP). Also, PaaS can be installed as a private cloud for free from [DigitalOcean marketplace](https://marketplace.digitalocean.com/apps/jelastic-paas/) running up to 10 nodes (Max 64 GB RAM each)
```
is that better? 
The amount of free resources depends on the provider and can be checked by the following link: https://jelastic.cloud/?freeResources=CLOUDLETS_STORAGE_NETWORK_IP
Free Lite Edition can be installed on Digital Ocean: https://marketplace.digitalocean.com/apps/jelastic-paas/

ripienaar: Doesnt it mean the jelastic part is still only free for 14 days? You might be managing free resouces but the manager is not free.  Jelastic is the manager so free or not depends on it alone, not other providers.

SlavaKatiukha: Every service providers offers Jelastic on its own rules - trial period length, further price per resource, free tiers, etc. After trial period, user needs to provide billing details in order to upgrade to full account. Further, if the consumption by the customer is not overcoming the limit of resources available for free, the services will be provided free of charge.

ripienaar: yeah so still, jelastic then isnt a saas on its own - its all the 3rd parties hosting / offering it and their terms?

I get it, but its just too complicated to really verify all this and make a honest representation to people about this service.

SlavaKatiukha: yes, it's not a SaaS but more PaaS on top of different cloud hosting providers. You can keep just information about free installation on DigitalOcean if it's simpler to verify

ripienaar: It's still not free then - since you have to pay DO.  I think this just does not qualify for ths list

SlavaKatiukha: Ok, thank you anyway!

PR #1586, opened 2020-07-21
Remove APIFull

abhay-ranawat:
APIFull Does not provide a free tier , they just provide a 1000 sign up credit for one time use
https://apifull.com/pricing.html


ripienaar:
thank you, they must have changed their offering, happens a lot!

PR #1588, opened 2020-07-22
Add cloudamqp.com

php-coder:
See https://www.cloudamqp.com/plans.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -716,6 +716,7 @@ Table of Contents
 
    * [airtable.com](https://airtable.com/) — Looks like a spreadsheet, but it's a relational database, unlimited bases, 1,200 rows/base and 1,000 API requests/month
    * [elephantsql.com](https://www.elephantsql.com/) — PostgreSQL as a service, 20 MB free
+   * [cloudamqp.com](https://www.cloudamqp.com/) — RabbitMQ as a Service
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -716,6 +716,7 @@ Table of Contents
 
    * [airtable.com](https://airtable.com/) — Looks like a spreadsheet, but it's a relational database, unlimited bases, 1,200 rows/base and 1,000 API requests/month
    * [elephantsql.com](https://www.elephantsql.com/) — PostgreSQL as a service, 20 MB free
+   * [cloudamqp.com](https://www.cloudamqp.com/) — RabbitMQ as a Service
```
ripienaar, line 719: maybe add a bit of dets about the free limits?

php-coder: I wouldn't mind and but I wasn't able to decide what to mention. In case of Postgres we mentioned the diskspace but what is most important for queues? They have (at least) the following limits:
- Max 100 queues
- Max 10 000 queued messages
- Max idle queue time 28 days
- 1M messages per month
- 20 Max Connections
Which one you think we should include here?

ripienaar: I think the most important is 1M messages and 20 max conns, maybe hehe

php-coder: Updated. Please, squash the commits during merge :)

ripienaar:
thank you

PR #1589, opened 2020-07-23
Update README.md

012345:
Added DataStax Astra under DBaaS to show 10GB free tier for Apache Cassandra backed Apps


ripienaar:
thanks

PR #1595, opened 2020-07-29
Add Scraper.AI

maximgeerinck:
Proposal to add scraper.ai to the list


ripienaar:
thank you :) for future ref, if you edit the same branch as a PR and push again additional commits gets added to the PR, that way you could have changed things around without making new PRs

However, new PRs also work, so as you wish :)

PR #1596, opened 2020-07-30
Update gke restrictions

toonsevrin:
https://cloud.google.com/kubernetes-engine/pricing only the first gke cluster is free (I think this changed this or last year).


ripienaar:
thank you!

PR #1622, opened 2020-08-28
Add Uploadcare Developer Hub

Skavr777:
In addition to paid services, Uploadcare has an absolutely free one for developers. It is called Uploadcare
Developer Hub There are a lot of Storage and Media Processing instruments, and I hope it would help someone.


ripienaar:
thank you

maple3142:
It seems that it requires people to fill a form and wait for them to verify, not just register and done. Does this qualify for this list?

ripienaar:
Yeah the signup seemed a bit complex but they are clear about what you get and doesnt seem to imply there are weird criteria for if you get it or not so I let it join the list, however if someone do sign up and reports that it's a nightmare I'd remove it

PR #1625, opened 2020-08-28
serveo.net seems dead

berdav:
Hi,
as title says, serveo.net seems unreachable.

If it is should be reachable but do not have site, please provide more information on how to connect.


ripienaar:
thank you

PR #1627, opened 2020-08-30
Update README.md

TxSadhu:
- Update Mailerlite with the new pricing model
- Added daily sending limit to mailjet.com


ripienaar:
thanks

PR #1633, opened 2020-09-05
Remove Migadu

louis-lau:
Migadu has killed their free plan a couple of weeks ago:
https://www.migadu.com/pricing/#what-happened-to-the-free-plan


ripienaar:
thank you

PR #1635, opened 2020-09-09
Added Zoho Forms in Form section

s-abinash:
Zoho form provides free form response service. Added that to the form category.


ripienaar:
thanks

PR #1646, opened 2020-09-13
Add TryCloudflare

maple3142:
Although Cloudflare is already there, but I think this is a different service similar to ngrok, but only support HTTP (?).

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,7 +585,12 @@ Table of Contents
 
   * [bootstrapcdn.com](https://www.bootstrapcdn.com/) — CDN for bootstrap, bootswatch and fontawesome.io
   * [cdnjs.com](https://cdnjs.com/) — CDN for JavaScript libraries, CSS libraries, SWF, images, etc.
-  * [cloudflare.com](https://www.cloudflare.com/) — Basic service is free, good for a blog, it also offers a free SSL certificate service and 5 firewall rules.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -752,6 +752,7 @@ Table of Contents
 
 ## STUN, WebRTC, Web Socket Servers and Other Routers
 
+   * [TryCloudflare](https://developers.cloudflare.com/argo-tunnel/trycloudflare) — Expose local HTTP servers through Argo Tunnel to public.
```
ripienaar, line 755: I'd prefer we just update the existing one with additional information

maple3142: But the existing CloudFlare entries are about CDN, DNS and Workers, which is quiet different from this feature, or you want to add CloudFlare into major cloud and group them together?

ripienaar:
While they arent a major cloud provider I do think they can be combined into 1 in the CDN section

maple3142:
You mean merge other Cloudflare products such as workers.dev into a single thing under CDN section?

ripienaar:
Yeah - workers is its own brand so we can leave that one alone, combining the rest is good - its not ideal I agree but the list is very long and we have a number of these cross purpose tools combined into one but some like CF remains

ripienaar:
thanks!

PR #1648, opened 2020-09-14
adds teletype to list of collaboration tools

akshaykmr:
I'm working on privacy focussed collaboration tools, would appreciate if teletype can be added to the list.
You can share terminals, voice, code, whiteboard and more. no sign-in required, end-to-end encrypted.


ripienaar:
thanks

PR #1650, opened 2020-09-14
Remove Engine Yard

maple3142:
Because there website doesn't mention anything about free 500 hours now, only trial.


ripienaar:
thanks!

PR #1651, opened 2020-09-14
Added links to image, VAT, avatar, and email API's

heymattbrown:


README.md
```diff
@@ -511,6 +511,7 @@ Table of Contents
 ## Email
 
   * [10minutemail](https://10minutemail.com) - Free, temporary email for testing.
+  * [Abstract Email Validator](https://www.abstractapi.com/email-verification-validation-api) - Email verifiction and validation API, with SMTP + MX checks
```
ripienaar, line 514: Please add something like `500 requests / month, non commercial use only`

Then in general we do not list the same thing in many catagories, I appreciate this is different APIs but its one company so we need to find a suitable place to list them and list it all in one entry

PR #1653, opened 2020-09-16
Update README.md

nsturtz:
Added Sturtz Network


README.md
```diff
@@ -697,6 +697,7 @@ Table of Contents
   * [raidboxes.io](https://raidboxes.io/) — Managed WordPress hosting with lots of included features. Free for developers and agencies, comes with custom domain support.
   * [readthedocs.org](https://readthedocs.org/) — Free documentation hosting with versioning, PDF generation and more
   * [render.com](https://render.com) — A unified platform to build and run all your apps and web app free SSL, a global CDN, private networks and auto deploys from Git, free for static web page.
+  * [sturtz.cf](http://sturtz.cf) - Create a free website! For Developers!
```
ripienaar, line 700: thank you, please add a few notes about the free tier - for example how much space, b/w and if its ad supported

maple3142:
The website looks weird, because its https returns a certificate from go.runhosting.com .

ripienaar:
these free websites are usually just rebrands of larger networks - to be fair though the link in the list is http not https

ripienaar:
But yes, on closer inspection the better solution would be to just add runhosting and not some rebranded shell around it.  So I'll close this one.

PR #1662, opened 2020-10-01
Update the Email list and UI and design list by adding free tools

mohammed786:
**Regression**
Update the **Email** list and **Design and UI** list by adding free tools

**Risk Impact**
Low


ripienaar:
thank you

PR #1663, opened 2020-10-01
Remove title link

schweikart:
There is a link in the title of the README.md document that breaks the generation of the left-side navigation bar in the docsify-rendered page at [free-for.dev](https://free-for.dev). It seems like docsify can't handle links in headings. I have removed the link in the document title and solved the rendering issue. You can find a "before-after" comparison attached below:

| With a link in the title | Without a link in the title |
| --- | --- |
| image | image |


ripienaar:
thanks, lets see if things improve

ripienaar:
yup, looks much better than you

PR #1665, opened 2020-10-01
Update some PaaS description

maple3142:
glitch.io used to unlimited, but now they limit free account to 1000 hours/month (Total).

Okteto Cloud's description is slightly confusing, because it will make you think that it is similar to other kubernetes services, but it is not.


ripienaar:
thank you

PR #1668, opened 2020-10-04
Added Bip

joelkennedy:
Added Bip to the Web Hosting section. Bip is a static web hosting provider, and offers a free plan with 100GB/month bandwidth.


ripienaar:
thank you

PR #1669, opened 2020-10-06
New services

sanarena:
hotjar was mentioned two times (in analytics and in session recording).
remove it from the analytics section and kept it in the session recording section.
howuku belongs to the session recording category, moved it from analytics section
Add reactflow to session recording category
Add hitsteps to analytics category
add pingmeter to monitoring service category
add ip.city to APIs

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,6 +461,7 @@ Table of Contents
 
 ## Monitoring
 
+  * [Pingmeter.com](https://pingmeter.com/) - 5 uptime monitors with 10 minutes interval. monitor SSH, HTTP, HTTPS, and any custom TCP ports.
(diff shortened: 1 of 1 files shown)


sanarena:
A note to add regarding hotjar/howuku category change.
Their main service is session recording, same as fullstory, smartlook, mouseflow, ...

Since smartlook/mouseflow/.. are in "session recording" section only, so rest must follow.
Otherwise it mouseflow/smartlook/... can be considered analytics too and have duplicate link in both section.

So I think it is better to keep analytics only for analytics focused services.

ripienaar:
yup agree on hotjar, we prefer things listed only once - think they might have been added early on before I realised quite how large this list would become hehe

thanks!

PR #1675, opened 2020-10-11
Add Panelbear Analytics

anthonynsimon:
Just adding Panelbear Analyics, which has a free tier with no time limit.

Amazing list BTW, this is incredibly useful for everyone! Thanks for your efforts :)


ripienaar:
thank you

PR #1694, opened 2020-11-03
added lambda store

rafaelnu:
https://lambda.store/ now supports TLS in free-tier.
Hopefully it can be added to list.


ripienaar:
thank you

PR #1695, opened 2020-11-03
Removed Dployr

mashuk999:
They no longer provide service for free. You can make a account but cannot create a instance for free, it requires a paid plan


ripienaar:
thanks

PR #1704, opened 2020-11-18
Clean up AWS link

logankuzyk:
The new link goes to the same place, and displays more free services when the page loads. It matches the other links on the page better.


ripienaar:
thank you

PR #1710, opened 2020-11-21
Update Fauna Amount

samrobbins85:
Fauna has updated their pricing https://fauna.com/pricing, so now the limit is much lower


ripienaar:
Thanks

PR #1717, opened 2020-11-27
Added Similar Words API

rvigneshw:


rvigneshw:
Not placed that in sorted manner.

PR #1731, opened 2020-12-08
Added web.dev

isuckatprogram:
Web.dev is a great tool that I found online! Other developers may be able to use it to improve their ranking on search engines!


ripienaar:
thank you

PR #1741, opened 2020-12-18
Remove heatlyanalytics

Admiral-Enigma:
As they have sunset their product, it makes sense to remove it from the list


ripienaar:
thank you

PR #1745, opened 2020-12-24
Added Protectumus Website Security Scanner

pascut:
The website was removed 22 days ago because of being down. I added again the website to this list because Protectumus have an uptime of 99,99%. Also, we added new security features that are available in the free plan.


ripienaar:
thank you

lodee77 (commented):
Ok

PR #1749, opened 2021-01-02
000webhost has ads on each pages

tomap:
See a quick test I did: https://tpi2-eu.000webhostapp.com/


ripienaar:
Thsnks

ripienaar:
Actually would just remove it tbh

tomap:
> Actually would just remove it tbh

You can :)

PR #1751, opened 2021-01-06
Adding Prepros.io to the IDE and Code Editing

aureq:
Adding Prepros.io to the IDE and Code Editing. Similar to VSCode but focused on web development. There is no functional difference between the free and paid version. The free version shows a dialog prompting you to upgrade at a random interval. You can buy a license to remove the dialog.


ripienaar:
Thnx

PR #1755, opened 2021-01-08
Renamed GitLab pages a more descriptive name

YoramGondwe:
>https://about.gitlab.com/product/pages/ was named gitlab.com and was not fully descriptive of the service, therefore, I changed it to GitLab Pages


ripienaar:
thanks

PR #1761, opened 2021-01-10
Add mailbluster

bitsnaps:
mailbluster is an email marketing software with free plan of 62k month and unlimited subscribers.


ripienaar:
thank you

abhay-ranawat:
Hey it’s now independent, it requires AWS account to send emails , so can it be included here ?
See here _“Yes, MailBluster sends and tracks emails through the user's Amazon account. MailBluster uses Amazon SES for sending and Amazon SNS for tracking emails.”_

bitsnaps:
> Hey it’s now independent, it requires AWS account to send emails , so can it be included here ?
I'm not sure if we can consider the free(mium) AWS account (I'm not using it) which I believe requires a valid visa card.

ripienaar:
Ah you're right, I missed this during review, I'll revert the PR

thank you so much @abhay-ranawat and @bitsnaps

PR #1763, opened 2021-01-12
Added unixtime API

rodenacker:
* [Unixtime](https://unixtime.co.za) - Free API to convert Unixtime to DateTime and vice versa.


ripienaar:
thanks

PR #1766, opened 2021-01-17
Updated postale.io mail entry

ReduxGB:
Postale.io changed their free plan recently, commit includes the updates to their plan.


ripienaar:
thank you

PR #1769, opened 2021-01-19
Formspree no longer requires exposing your email

daviddarnes:
Formspree now uses custom endpoints which can be created in your Formspree account without having to expose your email. More info here https://help.formspree.io/hc/en-us/articles/360056076314-Phasing-out-legacy-forms-email-URLs


ripienaar:
Thank you

PR #1772, opened 2021-01-20
Correct Localazy pricing information

honzabilek4:
I'm just fixing the misleading information on Localazy pricing. This should be more correct as per https://localazy.com/pricing


ripienaar:
Thank you. Often free tiers change. What once was unlimited becomes limited etc

PR #1779, opened 2021-01-22
docs: changes in gitlab ci free plan

kabaluyot:
The GitLab free minutes build time was changed from 2,000 minutes to 400 minutes last Sep 2020.

It's good to note also that the `usage quota` is measured per top-level group.

Please see this article: https://about.gitlab.com/releases/2020/09/01/ci-minutes-update-free-users/


ripienaar:
thanks

PR #1780, opened 2021-01-23
Use gihub actions

staeff:
Hi, I tried to move the awesomebot linkchecking from travis to github actions, because I hoped they might run faster there. They don't, it feels like they need about the same time :)
Maybe it is integrationwise still a win and you still want to merge this? If not thats also fine of course :)


ripienaar:
Honestly I don’t care for theee checks. I merge things before they are done most of the time. Happy to just delete them :)

staeff:
Fair enough :)

PR #1782, opened 2021-01-24
Add GitHub Actions

Tostisto:


ripienaar:
The CI/CD features of github is already listed in its existing listing, cant really realistically list these big multi faceted things under all possible categories - github would end up being listed 5 or 6 times. So not sure we should be adding this one.  We could perhaps expand the existing entry

Tostisto:
Oh, I didn't notice that GitHub CI/CD is already included in the Source Code Repos. But I think for clarity it might be easier to include it in the CI / CD.

ripienaar:
agree, it would be clearer, but prefer not to list them many times, the list is already pretty much too big to consume

Tostisto:
I understand, there is a lot

PR #1783, opened 2021-01-25
Update README.md

devnulldevzero:
Add a Azure always free web plan. 
Proof: https://azure.microsoft.com/ru-ru/free/

### DIFF
--- a/README.md
+++ b/README.md
@@ -708,7 +708,7 @@ Table of Contents
   * [Liteflow](https://liteflow.com/) - Low-code development toolkit built to help you focus on your app’s real value.
 
 ## Web Hosting
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -708,7 +708,7 @@ Table of Contents
   * [Liteflow](https://liteflow.com/) - Low-code development toolkit built to help you focus on your app’s real value.
 
 ## Web Hosting
-
+  * [Azure hosting](https://azure.microsoft.com/ru-ru/free/) - Zero cost web hosting, you can create up to ten web applications, based on PHP, Node.JS, etc. (there's in always free plan)
```
ripienaar, line 711: There is already an entire huge section dedicated to azure top of the document, dont think we want to duplicate anything elsewhere

devnulldevzero: hm.. if this manual have a huge information about Azure in one place, so, why you can't split it into all headers?

ripienaar: because we prefer not to list 1 thing in every possbiel catagory

PR #1802, opened 2021-02-14
Removing KintoHub

unixfox:
KintoHub is shutting down on the March 12th, 2021: https://www.kintohub.com/blog/2021/02/12/KintoHub-Team-Joins-Cash-App/


ripienaar:
thanks

PR #1803, opened 2021-02-18
Add logchimp product

mittalyashu:


ripienaar:
this doesnt seem like a SaaS to me, you can self host or install into another host - but as a thing itself its not a SaaS

PR #1806, opened 2021-02-19
Add notice to forwardemail.net

aboutDavid:
Recently I got an email from forwardemail.net that looks like this:

So, I think everyone should know about this because it isn't "Free for dev" if you use these TLDs


ripienaar:
nice, thanks

AR-Student824:
bruh thats just annoying

sldless:
ok

PR #1810, opened 2021-02-23
docs(log management): Add LogDNA

nimbinatus:
LogDNA has a free tier that's separate from the free trial.

Full disclosure: I'm a LogDNA employee.

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,6 +467,7 @@ Table of Contents
 
   * [bugfender.com](https://bugfender.com/) — Free up to 100k log lines/day with 24 hours retention
   * [humio.com](https://www.humio.com/) — Free up to 2 GB/day with 7 days retention
+  * [logdna.com](https://logdna.com) - Free for a single user, no retention, unlimited hosts and sources
   * [logentries.com](https://logentries.com/) — Free up to 5 GB/month with 7 days retention
   * [loggly.com](https://www.loggly.com/) — Free for a single user, see the lite option
   * [logz.io](https://logz.io/) — Free up to 3 GB/day, 3 days retention


README.md
```diff
@@ -467,6 +467,7 @@ Table of Contents
 
   * [bugfender.com](https://bugfender.com/) — Free up to 100k log lines/day with 24 hours retention
   * [humio.com](https://www.humio.com/) — Free up to 2 GB/day with 7 days retention
+  * [logdna.com](https://logdna.com) - Free for a single user, no retention, unlimited hosts and sources
```
ripienaar, line 470: does `no retention` mean just tail -f? As a free tier thats quite far from the paid service. The aim of this list is to let people working in this space get a full(ish?) feature set they can use to build/demo/etc and later move paying customrs over to the full service.  If `no retention` translates to tail -f only, that might be a bit light

nimbinatus: Very fair criticism. It's basically live tail for unlimited sources. You can pull in as many servers, clusters, and platforms as you'd like into the same tail view (and you can separate sources out to different views in any combination) that you can search; and there's no limit on the size you send regarding log lines per day.

I've been working on getting the community the retention part since I started, but I completely understand that the free tier offering might not quite be up to be here on this list just yet. I'm good with swinging back and reopening this PR once we either get our free tier set up with some retention or we bring stronger feature parity to our free tier.

ripienaar: OK, so lets merge this one instead, since you're working there and championing this I'll trust you to pop by later and update it should that become available

thank you

nimbinatus: Thanks! I'll be back to update :) And I'll add anything else I come across. Thanks for the cool resource!

PR #1817, opened 2021-02-23
Add Darklang

pbiggar:
Darklang is a pretty wide product and fits in a lot of the categories, so I tried to keep it in just the major categories where you would go looking for something like Dark: BaaS, PaaS and Editor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -665,6 +665,7 @@ Table of Contents
   * [appharbor.com](https://appharbor.com/) — A .Net PaaS that provides 1 free worker
   * [configure.it](https://www.configure.it/) — Mobile app development platform, free for 2 projects, limited features but no resource limits
   * [codenameone.com](https://www.codenameone.com/) — Open source, cross platform, mobile app development toolchain for Java/Kotlin developers. Free for commercial use with unlimited number of projects
+  * [darklang.com](https://darklang.com/) - Hosted editor/language/infrastructure
   * [Deta](https://www.deta.sh) – Deploy unlimited number of Node.js and Python apps for free up to 50k requests/month. Includes free DBs, Auth and email.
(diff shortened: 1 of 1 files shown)


ripienaar:
Thanks. We try to list things in one category only - the list is a bit wild already 

further I don’t see details about what, or even if, they wil offer a free tier post their betas which is a requirement for being on the list. 

Could have missed it though feel free to stock a link here with details

pbiggar:
Thanks for the reply.

> Thanks. We try to list things in one category only - the list is a bit wild already

OK, sounds good. It probably makes sense to stay in the BaaS category so I'll make that change.

> further I don’t see details about what, or even if, they wil offer a free tier post their betas which is a requirement for being on the list.

Yes, we haven't actually figured out what we'll offer or even what unit we'll offer in. Would something like "free during the beta, generous free tier planned after beta" work for you?

ripienaar:
Yes that would work - see top of the readme that’s about the level of rigor I apply, just want to know a free tier is planned

pbiggar:
> Yes that would work - see top of the readme that’s about the level of rigor I apply, just want to know a free tier is planned

Great, I've updated the PR. Thanks!

ripienaar:
Ah, well sorry I was not clear, I need to be able to see this stated on the website of the service in question, not in the comment here - sure keep it here of course, but I have to be able to verify the offer and also those who later update entries.

So if there's a place on the https://darklang.com/ site that says its planned great, else I just don't know.

Sorry for not being clearer earlier, I know updating this is a pain

pbiggar:
Gotcha! How's this? https://darklang.com/language#faq-pricing

ripienaar:
winner

ripienaar:
thank you  :)

pbiggar:
Awesome, thanks!

PR #1819, opened 2021-02-23
Add redirection.io

damienalexandre:
Add https://redirection.io/ in the Miscellaneous section.

There is a free plan: https://redirection.io/pricing


ripienaar:
thanks

PR #1820, opened 2021-02-23
Minor: Put "other" link at the bottom rather than between "Monitoring" and "PaaS"

remram44:
Alphabetical order is good but this one should be at the end.

Thanks!


ripienaar:
thanks

PR #1821, opened 2021-02-23
Update Mixpanel free plan (it has changed)

illiriks:
The Mixpanel free plan listed here has been discontinued for some time; the current free plan is much larger than the previous one. Source:  https://mixpanel.com/pricing/


ripienaar:
thank you

PR #1825, opened 2021-02-23
Updating Eclipse Che hosted instance details

l0rd:
che.openshift.io will be shut down the 1st of march 2021. workspaces.openshift.com is the new service that replaces it.


ripienaar:
thanks

PR #1836, opened 2021-02-24
Added wormhol.org

ricardbejarano:


ripienaar:
nice, been looking for a replacement for the old firefox send service.

PR #1838, opened 2021-02-24
Update Cloudinary's free tier details

daOrangePeeler:
Cloudinary's free forever tier has changed since it was first added. They now provides 25 monthly credits which can be used for 25K transformations, 25 GB of managed storage, or 25 GB of CDN usage.

Source: cloudinary.com/pricing


ripienaar:
thank you

PR #1843, opened 2021-02-24
Added Thundra to Monitoring

milkers:
Thundra has a generous free tier for application monitoring and debugging.


ripienaar:
nice, thanks

PR #1846, opened 2021-02-24
add updrafts.app

smeijer:
Updrafts.app is a drag and drop builder, similar to webflow. Hence the positioning next to webflow. The difference being, that updrafts.app is meant for developers that work with tailwindcss. They can design their pages for free in updrafts.app, and export the html + css to continue working on it in their editor.

There is a free plan available, that requires no credit-card, and let's the user create two projects, with both two pages, for non-commercial usage.


ripienaar:
Thank you

PR #1848, opened 2021-02-24
Add UVdesk under Tools for Teams and Collaboration

papnoisanjeev:
Add UVdesk

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,7 @@ Table of Contents
   * [wormhol.org](https://www.wormhol.org/) — Straightforward file sharing service. Share unlimited files up to 5GB to as many peers as you want.
   * [yammer.com](https://www.yammer.com/) — Private social network standalone or for MS Office 365. Free with a bit less admin tools and user management features
   * [zoom.us](https://zoom.us/) — Secure Video and Web conferencing, add-ons available. Free limited to 40 minutes
+  * [uvdesk.com](https://www.uvdesk.com/en/) — UVdesk is an online SAAS based customer support **helpdesk software** which allows customers to communicate over email, Social media with task management and many more other features as well.
 
 
 ## CMS


README.md
```diff
@@ -298,6 +298,7 @@ Table of Contents
   * [wormhol.org](https://www.wormhol.org/) — Straightforward file sharing service. Share unlimited files up to 5GB to as many peers as you want.
   * [yammer.com](https://www.yammer.com/) — Private social network standalone or for MS Office 365. Free with a bit less admin tools and user management features
   * [zoom.us](https://zoom.us/) — Secure Video and Web conferencing, add-ons available. Free limited to 40 minutes
+  * [uvdesk.com](https://www.uvdesk.com/en/) — UVdesk is an online SAAS based customer support **helpdesk software** which allows customers to communicate over email, Social media with task management and many more other features as well.
```
ripienaar, line 301: I believe the free tier is a open source self hosted product, not a service, so would not be suitable for this list, is that correct?

papnoisanjeev: UVdesk available with as a **SAAS** as well as **open source** self hosted product and both are different.
We requested here for **UVdesk SAAS** (https://uvdesk.com/en/member/signup) version.

ripienaar: I am not so sure, it seems the hosted version is paid.  The only free offering is download.  The free tier has to be SaaS.

Screenshot 2021-02-24 at 12 54 45

papnoisanjeev: Ok fine. Yes the **UVdesk SAAS** version is paid and subscription based.
Opensource version is free but It is not SAAS.

ripienaar: Then unfortunatley it would not be suitable for this list, sorry, thanks for the effort though I appreciate it.

PR #1851, opened 2021-02-24
Added netlifycms to CMS section

pdkn:


README.md
```diff
@@ -312,6 +312,8 @@ Table of Contents
   * [sanity.io](https://www.sanity.io/) – Hosted backend for structured content with customizable MIT licensed editor built with React. Unlimited projects. 3 users, 2 datasets, 500k API CDN requests, 5GB assets for free per project
   * [sensenet](https://sensenet.com) - API-first headless CMS providing enterprise-grade solutions for businesses of all size. The Developer plan provides 3 users, 500 content items, 3 built-in roles, 25+5 content types, fully accessible REST API, document preview generation and Office Online editing.
   * [GraphCMS](https://graphcms.com/) - Offers free tier for small projects. GraphQL first API. Move away from legacy solutions to the GraphQL native Headless CMS - and deliver omnichannel content API first.
+  * [NetlifyCMS](https://www.netlifycms.org/) - Open source content management for your Git workflow. Editor-friendly user interface. Intuitive workflow for content teams. Instant access without GitHub account
```
ripienaar, line 315: while yeah this is free as far as I understand there is no actual Service aspect to netlifycms - its a static site cms that you deploy on any other web host, including netlify.

So the thing thats a service in that scenario isnt the CMS, its the hosting.

pdkn: Surely it's a CMS service for static site generation? It's got a frontend, can hook into DAM services like Cloudinary etc. (it's free & open source)

ripienaar: Does the frontend run on servers offered as a service or its a local thing? 

So for example, hugo is a site CMS for static site generation, it's got much of what you mention and can run on many services - but hugo is not a service.  So it's not on the list.

I am trying to understand if this CMS is similar to Hugo or its a service on its own.

pdkn: aha, I missed the note in intro. It's similar to Hugo. However I do find this a strange delineation which would bias against  free self-hosted services for #JAMSTACK. These kinds of CMSs are a valid a solution to pick as others like Contentful etc. and would be of interest to "System Administrator, DevOps Practitioners, etc"? As you've highlighted there's no hard-fast rule, so maybe consider how to add these kinds of flavours? Your call 😉

ripienaar: Yeah agree its a grey line, but - you said it - "self-hosted services", those are by definition not SaaS.

We have a similar problem with IDEs, they clearly are not SaaS generally, but we chose to list them.  

I think ultimately there are better places for JAM Stack stuff, its a huge huge topic and could totally be a list of its own. Like this one https://github.com/automata/awesome-jamstack

So I'll pass on this PR I think, sorry about that, really appreciate the contribution effort though!

PR #1857, opened 2021-02-24
Remove Whales from the Docker-related list

chrismeller:
Whales is currently broken because the Slack webhook token built into a sub-image it runs in Docker is invalid (why does it include a Slack hook?). It also involves sending information about your codebase to their Heroku-hosted service for the "secret sauce", which is not well advertised and doesn't seem to make sense or be necessary for the functionality of the product. I personally don't believe we should be advertising this service.


ripienaar:
good call, thank you

PR #1858, opened 2021-02-24
Add Optic

acunniffe:
Optic is an open source project that makes documenting APIs and managing changes developer friendly

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ Table of Contents
   * [News API](https://newsapi.org) — Search news on the web with code, get JSON results. Developers get 3,000 queries free each month. 
   * [OCR.Space](https://ocr.space/) — An OCR API which parses image and pdf files returning the text results in JSON format. 25,000 requests per month free.
   * [OpenAPI3 Designer](https://openapidesigner.com/) — Visually create Open API 3 definitions for free.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -207,6 +207,7 @@ Table of Contents
   * [News API](https://newsapi.org) — Search news on the web with code, get JSON results. Developers get 3,000 queries free each month. 
   * [OCR.Space](https://ocr.space/) — An OCR API which parses image and pdf files returning the text results in JSON format. 25,000 requests per month free.
   * [OpenAPI3 Designer](https://openapidesigner.com/) — Visually create Open API 3 definitions for free.
+  * [Optic](https://useoptic.com/) — the easiest way to document and test your APIs. Optic uses development traffic to learn your APIs behavior, write your specification, and help you update the docs whenever your API changes.
```
ripienaar, line 210: Hmm, hard to tell, but this does not appear to be a service as such rather than something you build into your project right and host yourself?

acunniffe: Hey @ripienaar -- yeah the service we offer is the GitBot but that's free forever. At some point we'll be adding a pricing model for running API diffs against traffic from your real environments -- cloud version of the tool. 

No pricing on that, but local usage and normal CI integration is self-hosted and free forever.

ripienaar: Yeah then I am afraid as it stands thats a bit too minimal for this list I think

acunniffe: Ok, when we add a proper pricing page I'll come back. thanks for the feedback and for keeping the quality here high

PR #1859, opened 2021-02-24
Add Kreya

CommonGuy:
Kreya is an emerging gRPC GUI client to call and test gRPC APIs. Basically Postman, but for gRPC (which Postman doesn't support) instead of REST APIs.


README.md
```diff
@@ -198,6 +198,7 @@ Table of Contents
   * [BigDataCloud](https://www.bigdatacloud.com/) - Provides fast, accurate and free (Unlimited or up to 10K-50K/month) APIs for modern web like IP Geolocation, Reverse Geocoding, Networking Insights, Email and Phone Validation, Client Info and more.
   * [IPTrace](https://iptrace.io) — An embarrassingly simple API that provides reliable and useful IP geolocation data for your business.
   * [konghq.com/](https://konghq.com/) — API Marketplace and powerful tools for private and public APIs. With the free tier, some features are limited such as monitoring, alerting and support.
+  * [Kreya](https://kreya.app) — Free gRPC GUI client to call and test gRPC APIs. Can import gRPC APIs via server reflection.
```
ripienaar, line 201: bit of a tough call, as it's not a service but a downloaded client side tool. Though its probably enough of an IDE that I think it'd pass my exception for IDEs, so I'll add it.

PR #1861, opened 2021-02-24
Add Censys to the README

thehappydinoa:
Censys allows users to discover the devices, networks, and infrastructure on the Internet and monitor how it changes over time. There is a free tier for our Search Engine and API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,6 +420,7 @@ Table of Contents
   * [auth0.com](https://auth0.com/) — Hosted free for development SSO. Up to 2 social identity providers for closed-source projects.
   * [Authress](https://authress.io/) — Authentication login and access control, unlimited identity providers for any project. Facebook, Google, Twitter and more. First 1000 API calls are free.
   * [bitninja.io](https://bitninja.io/) — Botnet protection through a blacklist, free plan only reports limited information on each attack
+  * [Censys](https://censys.io/) - Censys allows users to discover the devices, networks, and infrastructure on the Internet and monitor how it changes over time.
   * [cloudsploit.com](https://cloudsploit.com/) — Amazon Web Services (AWS) security and compliance auditing and monitoring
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -420,6 +420,7 @@ Table of Contents
   * [auth0.com](https://auth0.com/) — Hosted free for development SSO. Up to 2 social identity providers for closed-source projects.
   * [Authress](https://authress.io/) — Authentication login and access control, unlimited identity providers for any project. Facebook, Google, Twitter and more. First 1000 API calls are free.
   * [bitninja.io](https://bitninja.io/) — Botnet protection through a blacklist, free plan only reports limited information on each attack
+  * [Censys](https://censys.io/) - Censys allows users to discover the devices, networks, and infrastructure on the Internet and monitor how it changes over time.
```
ripienaar, line 423: I can't find pricing info on the site that shows about a free tier, mind sharing a link to the details?

thehappydinoa: Yeah absolutely, here you go: https://support.censys.io/hc/en-us/articles/360038761891-Research-Access-to-Censys-Data

ripienaar: OK, I found that and I was afraid that's what you'd link.  I think that doesnt make the cut for the intent of this list, it's valuable for sure, but this list is for generic devs doing generic things, the pure focus on research, the need for impactfulness and intent to share just isnt in the spirit with the other things on this list.

So I think I'll pass on this one, appreciate the contribution though, pity we cant add it

thehappydinoa: Ah were you looking for our definition of free users here: https://censys.io/tos/. Any user who signs up for an account gets API access for free.

ripienaar: Sure, this list being about SaaS that offers free tiers - I basically need to be able to verify whats free and if it even is, and if its generally free (no, mail us for access, or specific quite tight requirements)

So looking for that kind of detail

PR #1865, opened 2021-02-25
Remove bintray as it is being deprecated

erenon:
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/


ripienaar:
thanks

PR #1871, opened 2021-02-26
Add Translation.io

MichaelHoste:


README.md
```diff
@@ -505,6 +505,7 @@ Table of Contents
   * [POEditor](https://poeditor.com/) — Free up to 1000 strings
   * [SimpleLocalize](https://simplelocalize.io/) - Free up to 100 translation keys, unlimited strings, unlimited languages, startup deals
   * [transifex.com](https://www.transifex.com/) — Free for Open Source
+  * [Translation.io](https://translation.io) - Free for Open Source
```
ripienaar, line 508: its worth mentioning in the description that this is rails focussed, thanks - also that signup as a OSS project requires an email to be sent and approval (we dont tend to accept such offers tbh)

MichaelHoste: It's [Rails](https://translation.io/rails) and [Laravel](https://translation.io/laravel) for now but we're going to add React/Angular/Vue soon and make it more "global". Do we still need me to mention Rails and Laravel and maybe extend the list later?

We are very responsive by email and an approval request takes less than a day. It doesn't seem to bother people since we already provide for several dozens Open Source projects.

Thank you for your feedback!

ripienaar: Yeah description is fine then.  

The constraint around join by approval isnt about your responsiveness abut about my ability to verify that you're accepting open source (meaning all open source, not what you decide to approve).  The opaqueness around the process makes it hard for me to verify the claim to be on this list.

MichaelHoste: > The opaqueness around the process

We just check that the project is Open Source (repository link) and matches the project that was already synchronized (they usually create it during the free trial that doesn't need any credit card)

Maybe our wording is too vague. What are other apps doing to accept open-source projects? Would a more structured form instead of an email could do the trick? The process will still be manual at some point as we need to flag the project.

If you're more confortable with a structured form, we could create it next week and resubmit.

Thank you for your feedback!

ripienaar: Generally I find you let people sign in with their github and if its a public repo it's elligable, perhaps with an additional check around the license, past that I think the thing people do is "free for public repos"

There's I think one or two other projects on here that require contacing them for signup and I accepted those because they offered is easy to abuse and quite expensive for them - for example packagecloud has a super generarous open source program where they will host gigabytes of packages and provide all the bandwidth for that.  Fair enough they need to be quite rigorous with the approvals so I accepted that.

MichaelHoste: Hi @ripienaar,

Thanks to your feedback we improved the onboarding for open source projects and we made the text on the pricing page more obvious.

No more email, no more validation.
Direct access to the free unlimited plan for open-source projects with public repository.

I hope that it now fits your guidelines.

----------

Screenshot 2021-03-03 at 15 32 11

------

Screenshot 2021-03-03 at 15 32 39

ripienaar: very nice, thank you

PR #1873, opened 2021-02-27
Add Pglet

FeodorFitsner:


README.md
```diff
@@ -953,6 +953,7 @@ Table of Contents
   * [Octopus.do](https://octopus.do) — Visual sitemap builder. Build your website structure in real-time and rapidly share it to collaborate with your team or clients.
   * [Pencil](https://github.com/evolus/pencil) - Open source design tool using Electron.
   * [pexels.com](https://www.pexels.com/) - Free stock photos for commercial use. Has free API that allows you to search photos by keywords.
+  * [Pglet](https://pglet.io) - The fastest way to build internal web apps in the language you already know.
```
ripienaar, line 956: this looks pretty nice, however its largely self hosted.  There's a managed install of it but I can't find any hint that its intended to be a free forever, or pay with free tier or something like this - without seeing a strong hint that a free tier is planned I can't add it here.

FeodorFitsner: Thanks for looking into that! So, as I understand having an official, live service with a "Free" plan would help to get into the list?

ripienaar: yeah, thats pretty much the purpose of the list - to list services delivered as a service with a free tier

FeodorFitsner: Got it! I'll close this PR for now and get back to it later when we launch. Keep up the great work!

PR #1874, opened 2021-02-27
Netlify Description Ambiguity Corrected

vishvofficial:
In the Phrase **Static  Site or App**, The word **app** might confuse the visitor by making them think of  Web Apps Like Flask, NodeJS, etc.
**But As Netlify is JAM Stack**, they only host static sites or static apps like Angular, Vue and other similar Client Side JS Frameworks.


ripienaar:
Thanks

PR #1875, opened 2021-02-27
Please, remove Alibaba Cloud, its "free trial" is false advertisement

alx-:
Take a look at these reviews, **2.1 stars** rating speaks volumes:

https://www.trustpilot.com/review/alibabacloud.com

I've had a very similar experience to some of those reviews. Worst customer experience/support I've ever seen in my life when dealing with cloud/hosting providers.

> they claim that they have a free trial, but once you sign up you get a message that you are not eligible for the trial... when you contact them they will tell you to send tons of documents including your ID or passport... and your bank statement

> I would give it a 0 if I could. Stay away. Words can not describe how bad they are. Their software is faulty: SMS does not come through, Credit cards do not debit, authorisations for Cards do not happen, their support people are clueless and tries to push blame onto others. I wasted 5 hours with their support - never again!

> I would give it a 0 star. Absolutely lousy, and worse, a thief that deduct money from your credit card and refuse to reverse it after complaint lodged in the rawest form - still pretend that they do not understand the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,13 +133,6 @@ Table of Contents
     * Log Analysis - 500MB of daily log
     * Full, detailed list - https://www.ibm.com/cloud/free/
 
-  * [Alibaba Cloud](https://www.alibabacloud.com/campaign/free-trial)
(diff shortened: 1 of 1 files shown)


ripienaar:
Thank you

PR #1877, opened 2021-02-27
Add RunJS

lukehaas:


README.md
```diff
@@ -1023,6 +1023,7 @@ Table of Contents
    * [Prepros](https://prepros.io/) - Prepros can compile Sass, Less, Stylus, Pug/Jade, Haml, Slim, CoffeeScript and TypeScript out of the box, reloads your browsers and makes it really easy to develop & test your websites so you can focus on making them perfect. You can also add your own tools with just a few clicks.
    * [repl.it](https://repl.it/) — A cloud coding environment for various program languages.
    * [stackblitz.com](https://stackblitz.com/) — Online VS Code IDE for Angular & React.
+   * [RunJS](https://runjs.app/) - A desktop JavaScript and TypeScript playground for Windows, macOS and Linux.
```
ripienaar, line 1026: Thank you, I am afraid this list is for SaaS, we only really make an exception for IDEs and this is not an IDE.

lukehaas: @ripienaar would you be open to reconsidering RunJS now?
It looks like you have some similar tools in this section, like BlueJ.

ripienaar: no, if there are similar tools they should be removed

ripienaar: BlueJ seems like a full IDE, are you saying RunJS is a full IDE?

lukehaas: BlueJ is designed for beginners to learn Java. It's used for learning and teaching, the same way RunJS is used for learning and teaching JavaScript.

ripienaar: would rather remove BlueJ then

lukehaas: Ok, what about the code editor angle? You have editors like Sublime Text and Brackets that aren't IDEs. RunJS can function as a code editor like these.

ripienaar: I am fine with those.  Please take the hint, your tool not being on this list is not the end of the world. There are 100s of awesome lists many are better places with more Js developers than here focus your efforts there

PR #1878, opened 2021-02-27
Update Contentful

djpowers:
Terms of free tier were revised to match current offering

Source: https://www.contentful.com/pricing/#space-limits


ripienaar:
Thanks

PR #1889, opened 2021-03-02
Added & Updated content (see desc)

thispsj:
Added drv.tw and updated user limit of Microsoft Teams


ripienaar:
thank you

thispsj:
Welcome .
@ripienaar sorry but I mistakingly wrote <del>50k</del> instead of **500k** users in Microsoft Teams desc. So will you correct that or I should send another PR for that ?

thispsj:
Nvm @ripienaar sent another PR with more free services that also fixes the above mentioned thing.
#1892

PR #1907, opened 2021-03-11
Add free subdomain description to freedns.afraid.org

rizdaprasetya:
Added more description to `freedns.afraid.org` about it also provide free subdomain. Which I personally have test/use over the years and works pretty well.


ripienaar:
thanks

PR #1910, opened 2021-03-12
Remove Splunk

lucaskjaero:
Hi! Splunk does offer a free installation, but it is self-hosted. Since this list excludes self-hosted offerings, I think it should be removed. [More context here](https://docs.splunk.com/Documentation/Splunk/8.1.2/Admin/MoreaboutSplunkFree).


ripienaar:
Thank you

PR #1916, opened 2021-03-13
Added Penpot

Orphail:
Penpot is an Open Source web based design and prototyping platform that works with open web standards (SVG)


ripienaar:
Thank you

PR #1919, opened 2021-03-15
Added CatchJS

larspars:


README.md
```diff
@@ -566,6 +566,7 @@ Table of Contents
 
 ## Crash and Exception Handling
 
+  * [CatchJS.com](https://catchjs.com/) - JavaScript error tracking with screenshots and click trails. Free for open source projects.
```
ripienaar, line 569: Unfortunately we generally do not accept sites where you have to contact them and apply to qualify for the free tier, it's not clear what the criteria is etc

```
As software developers, we have much gratitude to the open source software community, and therefore provide a free subscription to such projects. Qualifying projects get full access to the Development plan, worth $49 per month, for free. ```

most sites thats free for OSS simply take a GH login and check the associated repo is public and perhaps have a license

larspars: Thanks for checking out the PR.

What you're describing is actually how the open source plan for CatchJS has always worked, you give a github link and get an email back with an account set up. I'm the one who's been handling those requests, and no one has ever been rejected. I realize now that the text on the page talks about "qualifying projects" which might make it seem like there's more to it, so I've updated the description to reflect that there's no qualification other than having an open license. Thanks for this feedback :)

ripienaar: thank you

PR #1932, opened 2021-03-28
Add Mergify

jd:
Not 100% sure about the category, but I think it makes sense in CI/CD.


ripienaar:
thank you

rizqventure:
Issue moved to [ripienaar/choria-demo #1](https://github.com/ripienaar/choria-demo/issues/1) via [**ZenHub**](https://www.zenhub.com/)

PR #1941, opened 2021-04-06
Added freecurrencyapi.net

freecurrencyapi:
Free currency conversion and exchange rate data API. 10 requests/hour without an API key, 50 000 requests per month when you register for free.


ripienaar:
thanks

PR #1943, opened 2021-04-07
remove apimetrics.io

Dem0n13:
Can't find any free plan, only trial: https://apimetrics.io/apimetrics-pricing/


ripienaar:
thank you

PR #1948, opened 2021-04-08
Add a DNS Provider

smesgary:
Add ArvanCloud to DNS Providers


ripienaar:
We could perhaps expand them a bit as here?

Screenshot 2021-04-08 at 20 35 49

smesgary:
Yes, I think we can do it

thispsj:
@smesgary @ripienaar 
The entry is already expanded. I'd already done that in 7ad8371 commit. You can remove that entry from DNS section as we don't need multiple entries for the same site. Yes you can correct my mistake of ~~sites~~ to domains.

ripienaar:
ah fair, i didnt notice its a dupe, well spotted

PR #1955, opened 2021-04-14
Add zdoo.co and zentao.jp

PixelTroy:
Notice:
To open a free site on zentao.jp, you should click クラウド and register an account. And then click 申込み on https://zentao.jp/page/price.html, and choose the "Mini" version. It's completely free forever with limited users and space.


ripienaar:
Thank you, is there a english version of this page? We can only accept english submissions I am afraid

PixelTroy:
Then i guess zentao.jp is not qualified, but I guess zdoo.co is ok.

ripienaar:
Yes, please update the PR, thank you

PixelTroy:
I have removed zentao.jp, sorry for the delay.

PR #1960, opened 2021-04-18
Add Clouflare Pages

froblesmartin:


ripienaar:
We prefer to list things once rather than in every possible category, both of these already have entries please expand them instead

thanks!

ripienaar:
(closed by accident)

froblesmartin:
@ripienaar I think it would be clearer to have each specific product on each section (even though it would be the same vendor), but okay, I have removed GitHub Pages and moved Cloudflare Pages into the Cloudflare existing entry. :+1:

ripienaar:
Clearer yes, but much harder to maintain and grow the list massively - already its bordering on unusable imo

thank you!

froblesmartin:
> Clearer yes, but much harder to maintain and grow the list massively - already its bordering on unusable imo
> 
> thank you!

@ripienaar 

You are welcome! :smiley: 

Well, that's why categories and search tools are used :yum: If I look into the categories and go to "Web hosting" I will miss some as they are not in that category. That's why I have added "Web hosting" in the new entry for Cloudflare in the description, if someone uses crtl+f and looks for that they will find it.

PR #1962, opened 2021-04-21
feat(web-hosting): add *awesome-hosting*

dalisoft:
Adding web-hosting resources from https://github.com/dalisoft/awesome-hosting


ripienaar:
Appreciate this is topical, but ultiamtely this list is to list SaaS, not other lists. So I'd rather stick to that topic.

Thanks for the effort though.

PR #1965, opened 2021-04-25
Added favicon

georgebawesome:
Its main purpose is to help visitors locate your page easier when they have multiple tabs open.


ripienaar:
thank you

PR #1970, opened 2021-04-27
Add Format Express - update README.md

Baldrick44:


README.md
```diff
@@ -1249,7 +1249,8 @@ Table of Contents
   * [RandomKeygen](https://randomkeygen.com/) - A free mobile-friendly tool offers a variety of randomly generated keys and passwords you can use to secure any application, service or device.
   * [Cronhooks](https://cronhooks.io/) - Schedule one time or recurring webhooks using api and web app. Free plan allows 1 webhook schedule.
   * [Hook Relay](https://www.hookrelay.dev/) - Add webhook support to your app without the hassles: done-for-you queueing, retries with backoff, and logging. The free plan has 100 deliveries per day, 14-day retention, and 3 hook endpoints.
-
+  * [Format Express](https://www.format-express.dev) - Instant online formatter for JSON / XML / SQL.
```
ripienaar, line 1252: It's nice and all, but perhaps a bit minimal to qualify as a SaaS as such imo?

Baldrick44: It certainly not an SaaS. I proposed it because it's similar to other links of the Miscellaneous section (like the Base64 encoder), and there is no other formatter in the list. 
It's your call. Have a nice day.

ripienaar: Fair point about the base64 encoder, thanks for the contribution!

PR #1973, opened 2021-04-29
add privacy management section

markmichon:
Bearer is no longer an API monitoring tool, so this PR does a few things:
- Removes Bearer from the monitoring section
- Creates a new privacy management category for tools like consent and cookie management, gdpr compliance, etc.
- Adds a handful of services that fit the category.

To make things a bit easier, here are their pricing pages in order of appearance:
- https://www.bearer.sh/pricing
- https://www.osano.com/plans
- https://www.iubenda.com/en/pricing
- https://cookiefirst.com/#pricing
- https://www.ketch.com/packaging

Wasn't sure where to put the category itself, so I dropped it right before the misc section, but the ToC position is still alphabetical.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Table of Contents
    * [PaaS](#paas)
    * [Package Build System](#package-build-system)
    * [Payment and Billing Integration](#payment-and-billing-integration)
+   * [Privacy Management](#privacy-management)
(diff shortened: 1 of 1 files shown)


ripienaar:
great, thanks for this!

PR #1978, opened 2021-05-09
Added AUR

n0noob:
Added AUR for Arch


ripienaar:
Not really sure it qualifies tbh it’s just too specific and tied to the distro. I don’t consider it a SaaS really - it’s a SaaS but really part of the distro really. Same for the suse and fedora equivalents

n0noob:
Yes, it is for Arch and distros based on Arch. Even I am a bit doubtful about the comparison of the things AUR has to offer as compared to other sites mentioned in this section.
Feel free to close this PR in that case.

ripienaar:
Thank you, yeah I think I'll pass on this one, appreciate the effort though!

PR #1888, opened 2021-03-02
Add 20i

cyberfountain:
Add 20i to Hosting and cdn category.

PR #1990, opened 2021-05-25
Revert "Add 20i"

FatGrizzly:
Reverts ripienaar/free-for-dev#1888

I don't really see any free packages on their website, and their "Free CDN" is only applicable on their paid hosting plans


ripienaar:
Thanks - sadly these things often change

PR #1991, opened 2021-05-25
[README.md enhancement] add a back-to-top button for each section

kevinadhiguna:
Hi, I added a button for each section, that navigates back to the table of contents. It seems that it would be helpful if a reader is interested in more than one section but they are separated. So, the reader can go back to the table of contents and click another section.

Any feedback will be appreciated, thanks !


ripienaar:
great thank you :)

PR #1992, opened 2021-05-27
Updated Oracle Cloud Free Tier offering

crazyuploader:
Recently Oracle have added ARM based Cloud Computing, and they've included it in free tier as well. Also block volumes is now 200 GB in total.

More info at: [Oracle Cloud Free Tier](https://www.oracle.com/in/cloud/free/#always-free)


ripienaar:
thanks

PR #2002, opened 2021-06-06
Add Scraper's Proxy

sazzy4o:
Add Scraper's Proxy (https://scrapersproxy.com)
* 100 free requests per month
    * More available if you request from support
    * Only successful requests are counted


ripienaar:
thanks

PR #2014, opened 2021-06-20
Added FlashDrive.io as PaaS

flashdriveio:


FatGrizzly:
image
FlashDrive seems like a free-trial based platform,
> For a service to be eligible it has to offer a free tier and not just a free trial. If the free tier is time-bucketed it has to be for at least a year.

ripienaar:
Agree, that would not qualify for this list.

PR #2017, opened 2021-06-21
Added Calendly & Flutterflow

thispsj:
This PR adds Calendly and Flutterflow.

Also ripienaar , how can I contact you ?

rip@devco.net <- will this work ?


ripienaar:
nice thanks, yes that email is correct.

PR #2021, opened 2021-06-21
Added xsl.tel free cPanel shared hosting

anzepintar:


FatGrizzly:
This Hosting really seems suspicious, It requires me to verify my phone number(which I did, and filled up all my details correctly), and later it just tells that my account doesn't exist.

@ripienaar Can you give me a few days to investigate this host before adding it to the list?

FatGrizzly:
and their TrustPilot pages really seems suspicious(most reviews seems to be botted)
https://www.trustpilot.com/review/xsl.tel

anzepintar:
I agree, it looks suspicious, but i have been using it for the past year and it is working ok

FatGrizzly:
> I agree, it looks suspicious, but i have been using it for the past year and it is working ok

A host which uses a nulled/cracked version of cPanel without a valid license is not welcome to this list(well, let ripienaar decide)

ripienaar:
> A host which uses a nulled/cracked version of cPanel without a valid license is not welcome to this list(well, let ripienaar decide)

agree, out of curiousity how do we check thats what they do?

FatGrizzly:
> agree, out of curiousity how do we check thats what they do?
Running a simple subdomain scan(https://subdomainfinder.c99.nl/) on their website let me know their server hostname, then check their hostname with http://verify.cpanel.net/

ripienaar:
nice, well then we can deny it, thanks

PR #2027, opened 2021-06-25
Remove Byet

FatGrizzly:
Well, this is again a Fake/Scummy host which uses a old cPanel version(which is modified heavily)

image
https://verify.cpanel.net/app/verify?ip=cpanel.byethost.com


ripienaar:
thank you

PR #2032, opened 2021-07-02
Add Microsoft Clarity

ICTman1076:
I have been using MS Clarity for some time now, and it seems like a good quality piece of software. There doesn't seem to be any limits from my experience, and [the FAQ concurs](https://docs.microsoft.com/en-us/clarity/faq).


ripienaar:
nice

PR #2035, opened 2021-07-02
Repl.it --> Replit

hackermondev:
Replit changed their name from ``Repl.it`` to ``Replit``

Learn more [here](https://blog.replit.com/dotcom).


ripienaar:
thanks

PR #2045, opened 2021-07-11
Add mocko.dev

gabriel-pinheiro:
Mocko is a mocking platform that lets you create mocks in the cloud, proxy your API and choose which endpoints to mock. All of that with a free plan that works, with no limits that get in your way.


ripienaar:
nice, thanks

PR #2050, opened 2021-07-14
Add EmailJS

bitsnaps:
EmailJS is free service to securely send emails from client side (P.S. I have not yet tested).


ripienaar:
thanks

PR #2053, opened 2021-07-17
000WebHost should not be on this list

mrNo0b:
Worst free hosting... also probably will sell your data, should be considered as malware.


ripienaar:
Thanks

PR #2055, opened 2021-07-18
Deleted unusable Travis Config.

thispsj:
This deletes the [Travis](https://travis.ci) YAML Config as it no longer runs builds. The credits are exhausted.


ripienaar:
thanks

PR #2057, opened 2021-07-22
Updating free tier offer of HeroTofu

ziogas:
Now it offers 100 free submissions per month versus previous 50


ripienaar:
thanks

PR #2059, opened 2021-07-23
Updated new offering details of internxt

blaazetech:
Due to recent demand from free-for.dev users, internxt seems to reduce their offering from 10GB to 2GB.
But we as developers still support you Internxt. :smiley:


ripienaar:
thank you :)

PR #2061, opened 2021-07-24
pp.ua free domain

kouhe3:
free domain - pp.ua


ripienaar:
thanks

ICTman1076:
Tried this service, appears to want card details. Perhaps worth noting?

ripienaar:
Many services require card details for free services, it prevents abuse and keeps the service safe.  Did you try it?

ICTman1076:
I did not register the domain in the end because I was uncomfortable with giving my card details, especially since it is a subdomain of pp.ua and not a FQDN.

PR #2071, opened 2021-08-03
Add GitLocalize

ilyaspiridonov:
Added GitLocalize (localization management platform for GitHub projects) under Translation Management


ripienaar:
thanks

PR #2073, opened 2021-08-04
Update description of Apify in README.md

fnesveda:
We've slightly updated our free plan (and our platform offerings in general), so I've updated the description of Apify to match the current situation.


ripienaar:
thank you

PR #2074, opened 2021-08-04
Add remote desktop tools section

faridshamsutdinov:
Add Getscreen.me
Move Zoho Assist

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Table of Contents
    * [Web Hosting](#web-hosting)
    * [Commenting Platforms](#commenting-platforms)
    * [Browser based hardware emulation](#browser-based-hardware-emulation-written-in-javascript)
+   * [Remote Desktop Tools](#remote-desktop-tools)
    * [Other Free Resources](#other-free-resources)
 
 ## Major Cloud Providers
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1393,6 +1394,13 @@ Table of Contents
 
 **[⬆ back to top](#table-of-contents)**
 
+## Remote Desktop Tools
+
+  * [Zoho Assist](https://zoho.com/assist) — Free for 1 remote technician & 5 unattended computers
+  * [Getscreen.me](https://getscreen.me) —  Free for 2 devices, no limits on the number and duration of sessions
```
ripienaar, line 1400: Generally we avoid listing the same service in multiple categories, there is already a huge section dedicated to zoho, if anything assist should move in there rather than add a 3rd place zoho is listed.

That leaves this desktop category with just 1 entry and we dont accept 1 entry-only categories.  So maybe we can find a few more free screen sharing solutions?

faridshamsutdinov:
@ripienaar 

> Generally we avoid listing the same service in multiple categories, there is already a huge section dedicated to zoho, if anything assist should move in there rather than add a 3rd place zoho is listed.

fixed

> That leaves this desktop category with just 1 entry and we dont accept 1 entry-only categories. So maybe we can find a few more free screen sharing solutions?

I added another Apache Guacamole™, which is what I used before. I will add more as I use it.

ripienaar:
thanks

PR #2095, opened 2021-09-02
update stackblitz description & add shortlink

rizdaprasetya:
Stackblitz has changed so much since it was 1st added. I'm updating the description of it to reflect current state.


ripienaar:
thank you

PR #2113, opened 2021-09-14
Google free tier changes

iamLiquidX:
the free compute engine instance is changed from f1-micro to e2-micro


ripienaar:
thank you

PR #2127, opened 2021-09-28
Add Duply.co

distartin:
Tool to create dynamic banners automatically. Can use API or just use URL directly


ripienaar:
thank you

PR #2129, opened 2021-09-30
Update Hasura's offering

lilnasy:
Hasura seems to have moved away from IaaS ([web.archive.org, 2016](https://web.archive.org/web/20160206232509/http://hasura.io/pricing)) to a GraphQL front-end for your existing database ([hasura.io](https://hasura.io/pricing/#faq))


ripienaar:
thank you

PR #2130, opened 2021-09-30
Remove blockspring.com: no free tier

lilnasy:
Can't find information about their former offering, but they only do 14-day trials now.


ripienaar:
thanks

PR #2163, opened 2021-10-12
add BlobCDN

rizdaprasetya:
Details as shown on the file diff.
In spirit of **Hacktoberfest, if you think this PR is valid** and accepted, please kindly add this label to this PR: `hacktoberfest-accepted`. Thank you 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -818,6 +818,7 @@ Table of Contents
     - Free Cloud DNS for unlimited domains.
     - Free Cloud Security with Basic DDoS Protection + 5 Firewall Rules.
     - Free VoD (Video On Demand) Platform with 10 GB Storage + 50 GB Traffic.
+  * [BlobCDN](https://blobcdn.com/) - Simple CDN for generating SVG shapes on the fly, can be customized via the url parameters.
   * [bootstrapcdn.com](https://www.bootstrapcdn.com/) — CDN for bootstrap, bootswatch and fontawesome.io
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -818,6 +818,7 @@ Table of Contents
     - Free Cloud DNS for unlimited domains.
     - Free Cloud Security with Basic DDoS Protection + 5 Firewall Rules.
     - Free VoD (Video On Demand) Platform with 10 GB Storage + 50 GB Traffic.
+  * [BlobCDN](https://blobcdn.com/) - Simple CDN for generating SVG shapes on the fly, can be customized via the url parameters.
```
ripienaar, line 821: not convinced I see the value in this one, bit far out of the intended audience for this list, what would one use this for?

rizdaprasetya: Sample use case:
1. Avatar placeholder/generator as a service. No need to locally store/generate avatar pic, just create an `img` html tag with src: `https://blobcdn.com/blob.svg?seed=<USER_ID_GOES_HERE>&size=30`.  Each user ID will produce different styled blob, but will be consistent.

 User Alice (id=101)
 User Bob (id=102)
 User Charlie (id=103)

2. "Blob" style element on web-frontend [was popular sometimes ago](https://speckyboy.com/blob-animated-backgrounds/), so instead of manually writing SVG blobs for your web's frontend background/element, just let this service generate random (or customized) one for you.

I understand your concern though, this "SVG-blob as a service" looks underwhelming, so feel free to accept/reject this PR as you see fits.

PR #2179, opened 2021-10-30
Remove netdepot

AliciaBytes:
https://netdepot.com/cdn/ only leads to a 404 page now and I can't find a free offering anymore.


ripienaar:
thank you

PR #2186, opened 2021-11-04
Make Selectel more clear

oddmario:
I thought it'd be worth mentioning that Selectel allows you to create unlimited DNS records for free which isn't common to be honest.


ripienaar:
thank you

PR #2196, opened 2021-11-10
Add Codepng

henripar:
Added Codepng - Codepng allows you to create awesome snapshots from your source code that you can share on social media. Choose between beautiful hand-selected background gradients and editor themes before exporting your code to PNG images.


ripienaar:
Not sure why any one thinks source code that can not be copy/pasted is "sharing" but ok :)

PR #2200, opened 2021-11-18
Update a little wrong info

harryawk:
I think this one is 100 MB instead of 1000 MB

I checked it on their site at 2021-11-18 04:20 AM GMT

image


ripienaar:
thank you

PR #2209, opened 2021-11-28
Adding design inspiration resources

kvenn:
Mobile Patterns, UI Garage, Mobbin, scrnshts, and dribbble

There weren't many resources in the design inspiration category so I added a few. All of which are free and somewhat different (and with different catalogs of content).

If you'd like me to remove any, I'd be happy to do so. I took the description from their pages directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1222,6 +1222,7 @@ Table of Contents
 **[⬆ back to top](#table-of-contents)**
 
 ## Design Inspiration
+
   * [awwwards.](https://www.awwwards.com/) - [Top websites] A showcases of all the best designed websites (voted on by designers).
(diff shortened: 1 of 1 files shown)


ripienaar:
Seems fine, you could also add a category for them since we now have quite a few?

Let me know if you rather not and I'll merge, thanks

kvenn:
Just created a new section and added a few more good, free resources. Let me know if this is what you had in mind!

README.md
```diff
@@ -1219,6 +1221,18 @@ Table of Contents
 
 **[⬆ back to top](#table-of-contents)**
 
+## Design Inspiration
```
ripienaar, line 1224: Very good, please add a blank line below the heading here and we're good to go! thanks for the great improvement and new entries

ripienaar:
thank you!

PR #2210, opened 2021-12-01
Update Render free tier information 

dnilasor:
This PR updates the Render entry in Web Hosting to include new free web services and databases and tweaks the grammar slightly.


ripienaar:
thanks

PR #2213, opened 2021-12-03
add /meet for Slack

adamcharvat:
/meet for Slack lets you generate Google Meet calls from Slack by using /meet


ripienaar:
thanks

PR #2216, opened 2021-12-06
Added squadcast to the list of monitoring tools that have a free plan.

muthali:
Added squadcast.com to the list of monitoring tools. Squadcast has a free forever plan for up to 10 users.


ripienaar:
thanks

PR #2219, opened 2021-12-08
Remove Appflux

AriESQ:
I ran a dead link checker on the free-for-dev page.

Removing Appflux as they do not appear to be around anymore.

Domain does not resolve. Little to no Twitter or Github traffic for several years.


ripienaar:
thank you

PR #2223, opened 2021-12-08
Remove vattly.com

AriESQ:
I ran a dead-link-checker on free-for-dev

Website does not resolve. No google or Twitter results.


AriESQ:
Did not pull master before additional commits. Closing PR.

PR #2226, opened 2021-12-08
Remove vattly.com

AriESQ:
I ran a dead-link-checker on free-for-dev

Website does not resolve. No google or Twitter results.


ripienaar:
Thanks for all of these, much appreciated!

AriESQ:
@ripienaar Do you prefer I do removals as individual PRs? Or would you prefer 1 PR where the proposed deletions are seperate/individual commits?

I'm not sure what Git best practice is for that scenario.

ripienaar:
1 would be great that would avoid merge conflicts and similar problems

PR #2232, opened 2021-12-13
Add mailmenot.io temp email service

KnightCrawlerDev:
Added http://mailmenot.io temp disposable email service to the Email section.


ripienaar:
thank you

PR #2238, opened 2021-12-27
Remove scraperapi.com

staeff:
The site does not have a free tier anymore, just a 5000 requests in 7 days trial. https://www.scraperapi.com/pricing/


ripienaar:
thank you

PR #2244, opened 2022-01-02
Update free plans offered by session recording tools

moran-shay:


ripienaar:
Wow, for once free services increasing their offering not decreasing, thank you

PR #2254, opened 2022-01-14
Removing https://staroid.com

brianredbeard:
staroid.com is no longer online. Non-SSL queries return a 404 from
Google Cloud Platform.


ripienaar:
thank you

PR #2255, opened 2022-01-14
Remove developer.suse.com

brianredbeard:
developer.suse.com now redirects to community.suse.com.  No mention of
services available to developers and appears more focused as a community
collaboration site.


ripienaar:
Also cant find it, but there's a conflict here we need to fix, suspect we'll have a few conflicts today :(

PR #2256, opened 2022-01-14
Remove drona.hq

brianredbeard:
In 2021 Drona [updated their pricing][1] to usage based.  At some point
they also [removed][2] the ability to "publish" apps from their developer
plan:

> App cannot be published and accessed by end users.

[1]: https://www.dronahq.com/usage-based-pricing/
[2]: https://www.dronahq.com/pricing/


ripienaar:
thank you

PR #2260, opened 2022-01-16
remove deploy now because they don't offer any free tier any more

mabujaber:


ripienaar:
Screenshot 2022-01-17 at 08 20 31

ripienaar:
So think I'l lleave it on, they dont make it obvious but its there

PR #2263, opened 2022-01-16
adding tabler icons under smart mockups

angelosds:
Tabler icons is a intuitive copy and paste icons platform for designers and developers


ripienaar:
thanks

PR #2265, opened 2022-01-17
Add Bruzu.com

motyar:
Bruzu is an API to generate images on the fly.


sriganesh:
Is there a free plan for this service?

motyar:
Yes you get 100 free API credit when you signup.
Also API is absolutely FREE with watermark.

ripienaar:
Are those credits one off or monthly? One off would not qualify for the list. See the top of the list.

motyar:
One off!
But api is unlimited FREE with watermark on images.

ripienaar:
OK, in that case update the readme entry with something indicating there's a watermark while free

motyar:
Done check https://github.com/ripienaar/free-for-dev/pull/2270

ripienaar:
thank you

PR #2275, opened 2022-01-17
Add Codiga

juli1:
Coding has an automated Coding Assistant, added under `IDE`.


ripienaar:
We generally list things just once, see if there's one catagory that would work for it? The list is already huge, listing same company twice and things get a bit crazy

juli1:
@ripienaar removed the `Code Quality` part then.

ripienaar:
thank you

PR #2284, opened 2022-01-20
Added Authgear in Security and PKI

fungc-io:
Added Authgear, an auth-as-a-service for websites and apps

Thanks you!


ripienaar:
thank you

PR #2289, opened 2022-01-21
Revert 2103 master

weber2017:


ripienaar:
What are you trying to achieve @weber2017 ?

PR #2291, opened 2022-01-21
Revert "Added formpost.app to forms"

weber2017:
Reverts ripienaar/free-for-dev#2287


ripienaar:
Why do you want to revert this?

PR #2295, opened 2022-01-22
Added psono.com

chickahoona:
Psono is a password manager for teams. It's community edition is completely free and the enterprise edition is free for up to 10 users. There is also a free hosted version for those who don't want to host Psono themself available under https://psono.pw

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,7 @@ Table of Contents
   * [Nuclino](https://www.nuclino.com) - A lightweight and collaborative wiki for all your team's knowledge, docs, and notes. Free plan with all essential features, up to 50 items, 5GB total storage.
   * [PageShare.dev](https://www.pageshare.dev) - Adds visual reviews capabilities into GitHub Pull Requests with no need to deploy websites. Free for up to 10 pages each month and 100MB of storage in total.
   * [Pendulums](https://pendulums.io/) - Pendulums is a free time tracking tool which helps you to manage your time in a better manner with an easy to use interface and useful statistics.
+  * [Psono](https://psono.com/) - Password manager for teams to securely store and share passwords and other secrets encrypted. Supports Windows, Linux, Mac, iOS and Android.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -345,6 +345,7 @@ Table of Contents
   * [Nuclino](https://www.nuclino.com) - A lightweight and collaborative wiki for all your team's knowledge, docs, and notes. Free plan with all essential features, up to 50 items, 5GB total storage.
   * [PageShare.dev](https://www.pageshare.dev) - Adds visual reviews capabilities into GitHub Pull Requests with no need to deploy websites. Free for up to 10 pages each month and 100MB of storage in total.
   * [Pendulums](https://pendulums.io/) - Pendulums is a free time tracking tool which helps you to manage your time in a better manner with an easy to use interface and useful statistics.
+  * [Psono](https://psono.com/) - Password manager for teams to securely store and share passwords and other secrets encrypted. Supports Windows, Linux, Mac, iOS and Android.
```
ripienaar, line 348: I am not seeing dtails of this hosted version you mention - like I see its there but whats its limits, guarantees, data management policies and more? Is it a service one can use and rely on?

chickahoona: Click at the top right on the "Free online demo" and you will land on https://psono.pw/ There are no feature limitations. The privacy policy can be found here https://www.psono.pw/privacy-policy.html

Yes it for sure is a service that you can use and rely on. Business model wise we are selling an enterprise version too and have a SaaS offer for companies to offer Single Sign On and other business feature. That's where the money comes from ;)

ripienaar: To me this just looks like a demo. Not something someone should use to store passwords (the hosted demo you mention)

if the free service had a description of whatever data commitment there is etc I would accept it. But this is just a nebulous unspecified demo and that’s just not really what this list is about

chickahoona: We don't advertise it as a production version as we try to keep costs low here, yet ~1300 ppl or so use it at the moment regularly. Anyway thanks for your consideration!

PR #2296, opened 2022-01-22
Adds Giscus.app to commenting platforms

adityatelange:
A comments system powered by GitHub Discussions. Let visitors leave comments and reactions on your website via GitHub! Heavily inspired by utterances.

https://github.com/giscus/giscus


ripienaar:
Nice but I don’t think it really qualifies for the list. The free SaaS here is GitHub.

adityatelange:
> Nice but I don’t think it really qualifies for the list. The free SaaS here is GitHub.

Yes, utterences mentioned in the list is also based on similar concept. Wdyt?

ripienaar:
Should also be removed in that case :)

PR #2300, opened 2022-01-24
fixes nootiz-link

ivanhofer:
I just saw that my previous PR contained an incorrect link. This fixes that issue


ripienaar:
thank you

PR #2303, opened 2022-01-26
Updates regarding PAAS Subscriptions

ErSauravAdhikari:
Suse has now turned off its developer platform. 

Divio has introduced new policies, which require you to now add a card and remove its previously promised support of custom domain names.

Post Regarding Divio: 
Require Card - https://www.divio.com/blog/divio-reinforces-its-crypto-mining-abuse-counter-measures/
Removal of Domain Support - https://www.divio.com/pricing/

As for SUSE: Their website for the SUSE developer platform no longer exists, it redirects to their community website.


ripienaar:
thank you

PR #2304, opened 2022-01-26
Update Commenting Systems, CI/CD, Dev blogs

dmitry-konchalenkov:

PR #2307, opened 2022-01-26
Revert "Update Commenting Systems, CI/CD, Dev blogs"

FatGrizzly:
Reverts ripienaar/free-for-dev#2304
TeamCity Professional is a self hosted software and doesn't qualify for the list.

image


ripienaar:
doh thanks for catching that

PR #2308, opened 2022-01-26
Optimize format and Correct outdated information

DreamOfIce:
Optimize the format, move fosshost and storj to IaaS, correct the CDN traffic of Arvan Cloud to 100GB / m, and add qingcloud DNS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -857,7 +857,7 @@ Table of Contents
 ## CDN and Protection
 
   * [Arvan Cloud](https://arvancloud.com/) — Offers cloud related services (CDN,Cloud DNS, PaaS, Security etc.). Free plan offers :
-    - CDN with Free SSL. 50 GB Traffic + 1 Million HTTP(S) Requests.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -893,24 +893,24 @@ Table of Contents
 
 ## PaaS
 
-  * [anvil.works](https://anvil.works) - Web app development with nothing but Python. Free tier with unlimited apps.
-  * [appharbor.com](https://appharbor.com/) — A .Net PaaS that provides 1 free worker
+  * [anvil.works](https://anvil.works/) - Web app development with nothing but Python. Free tier with unlimited apps.
+  * [appharbor](https://appharbor.com/) — A .Net PaaS that provides 1 free worker
```
ripienaar, line 897: lots of good changes here but I'd rather not remove the TLD like here - especially as there's a mix and match only some being added or removed, so would rather just leave as it and keep to fixing the content

thanks!

DreamOfIce: OK I'll delete those changes

PR #2311, opened 2022-01-27
Update AWS and light formatting

Hibot12:
Event though the EC2 egress expansion does not appear on their Free Tier website, [it indeed was updated](https://aws.amazon.com/blogs/aws/aws-free-tier-data-transfer-expansion-100-gb-from-regions-and-1-tb-from-amazon-cloudfront-per-month/) on November 2021.


ripienaar:
thanks

PR #2314, opened 2022-01-29
Added quackr.io to 'other free resources' section

p-michaels-20200:
Added quackr.io which is a free temporary and virtual phone number platform.


README.md
```diff
@@ -1556,6 +1556,7 @@ Table of Contents
   * [is-a.dev](https://www.is-a.dev/) — Free `*.is-a.dev` subdomain for developers.
   * [Microsoft 365 Developer Program](https://developer.microsoft.com/microsoft-365/dev-program) — Get a free sandbox, tools, and other resources you need to build solutions for the Microsoft 365 platform. The subscription is a 90-day [Microsoft 365 E5 Subscription](https://www.microsoft.com/microsoft-365/enterprise/e5) (Windows excluded) which is renewable. It is renewed if you're active in development(measured using telemetry data & algorithms).
   * [RedHat for Developers](https://developers.redhat.com) — Free access to Red Hat products including RHEL,OpenShift,CodeReady etc exclusively for developers. Individual plan only. Free e-Books also offered for reference.
+  * [quackr.io](https://quackr.io) — A free temporary and virtual phone number platform.
```
ripienaar, line 1559: Not sure its the kind of service the list is for, appreciate the add though but think I will pass.  I suspect we have 1 or two similar atm but would rather remove them than add more

PR #2327, opened 2022-02-14
Update README.md: delete Treescale 

zBugi:
delete Treescale because their service got shutdown since Dec 2021


ripienaar:
thank you

PR #2333, opened 2022-02-19
Add UptimeDog.com

MuratTopcu:
Free uptime monitoring service. Monitors availability of your website every 2 minutes 24/7/365

PR #2334, opened 2022-02-21
Revert "Add UptimeDog.com"

FatGrizzly:
Reverts ripienaar/free-for-dev#2333
Signup doesn't work.
image
@MuratTopcu


MuratTopcu:
Sorry, I added it because logging in is working. I'm an old user, I didn't try to add a new user :(

ripienaar:
I should probably start judging services even more by their cover. It was pretty clear by just looking at it this is not a reliable/quality service. But strictly speaking it fit.

ripienaar:
Thanks

PR #2337, opened 2022-02-22
Update draw.io to diagrams.net

staeff:
the product has been re-branded and the original link is redirected to diagram.net


ripienaar:
thanks

PR #2354, opened 2022-02-28
Add teleretro.com

tnyuan:
Remote retrospective tool, free plan includes 3 retros and unlimited members. Retro boards can be reused after creation.
https://www.teleretro.com/pricing

### DIFF
--- a/README.md
+++ b/README.md
@@ -1157,6 +1157,7 @@ Table of Contents
    * [taskade.com](https://www.taskade.com/) — Real-time collaborative task lists and outlines for teams
    * [taskulu.com](https://taskulu.com/) — Role based project management. Free up to 5 users. Integration with GitHub/Trello/Dropbox/Google Drive
    * [teamwork.com](https://teamwork.com/) — Project management & Team Chat. Free for 5 users and 2 projects. Premium plans available.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1157,6 +1157,7 @@ Table of Contents
    * [taskade.com](https://www.taskade.com/) — Real-time collaborative task lists and outlines for teams
    * [taskulu.com](https://taskulu.com/) — Role based project management. Free up to 5 users. Integration with GitHub/Trello/Dropbox/Google Drive
    * [teamwork.com](https://teamwork.com/) — Project management & Team Chat. Free for 5 users and 2 projects. Premium plans available.
+   * [teleretro.com](https://www.teleretro.com/) — Simple and fun retrospective tool with icebreakers, gifs and emojis. Free plan includes 3 retros and unlimited members.
```
ripienaar, line 1160: These 3 retros, I assume you configure a retro and then it can run for many months? It's not like you can use it 3 times or free?

tnyuan: There is no time limit on when you can run a retro again on the same board. You can use the same board as many times as you need.

ripienaar: thanks for clarifying

tnyuan: No problem, thank you! 👍

PR #2363, opened 2022-03-09
Adding UseCSV by Layercode to free-for.dev

rana8hesham:
Hi there, I would like to add UseCSV by Layercode to free-for.dev in other free resources. UseCSV is a CSV import functionality that users can add to their app in minutes. We offer a free plan that users can create unlimited Importers and upload files up to 100Mb.


ripienaar:
thanks

PR #2373, opened 2022-03-16
Added Nicelydone.club as design inspiration source

luminous8:


ripienaar:
It's not really 19k+ designs and there is no free tier. You can view a absolutely minimal set of design ideas without signing up - and there is no free tier.

Dont think it's suitable for the list sorry.

PR #2378, opened 2022-03-20
Add Pantry to "Storage and Media Processing"

imRohan:
Pantry is a free JSON storage API. We've built a very straightforward and easy-to-use API that you can use to CRUD your Pantry's data. We also offer an interactive dashboard that you can use to interact with your Pantry's baskets. This intuitive interface is especially helpful when debugging your application! This project is open-source and supported by a wonderful group of developers who continue to improve the platform.


ripienaar:
Seems nice but also seems without docs. Like one imagine tbe “jump right in” section would link to the libraries or something. 

I can’t eval this without signing up. I am not signing up for 100s or thousands of projects to evaluate if they are what they claim to be.

ripienaar:
Ah found the libraries.

PR #2379, opened 2022-03-21
Add Translized to Translation Management

tomovicn:


README.md
```diff
@@ -649,6 +649,7 @@ Table of Contents
   * [Texterify](https://texterify.com/) - Free for a single user
   * [transifex.com](https://www.transifex.com/) — Free for Open Source
   * [Translation.io](https://translation.io) - Free for Open Source
+  * [Translized](https://translized.com) - Free up to 1000 strings, 1 user, unlimited languages, unlimited projects
```
ripienaar, line 652: Specifically providing no support not even email support to free users is surely a bit of a joke?

Regardless, will merge, but these low quality service offerings are likely to be culled when I get some time to go through things

tomovicn: That's an error, thanks for pointing it out.
Obviously, we provide support for all users, we'll update that.
Thanks

ripienaar: thank you

PR #2381, opened 2022-03-21
Add is-a.dev subdomains

lucidwave:


ripienaar:
It’s already listed. You could perhaps love it to this more appropriate location from where it is now though

lucidwave:
yeah, I do agree that the domains section is a more proper place

PR #2391, opened 2022-04-06
Update gitlab pricing

Riezebos:
Gitlab is changing its free tier:
https://about.gitlab.com/blog/2022/03/24/efficient-free-tier/


ripienaar:
thanks

PR #2402, opened 2022-04-13
added railway also to the dbaas section

hangman1998:
Hi.
I saw that railway was in the PaaS section but **not** in the DBaaS section (unlike Heroku) and thought I should add it.

image


ripienaar:
We only list things once unfortunately so they have to go in tbe most appropriate section

hangman1998:
Oh! but Heroku is in both the aforementioned sections.

ripienaar:
I am sure we can find a exception to the rule left over from the early days :) but the right thing to do then is to remove heroku from more than 1 place or create a section in the major providers part if it has enough pieces in the offering to make it a major offering.

PR #2409, opened 2022-04-21
add domain js.cool

willin:


ChristianHeimke:
The link is broken - I guess the correct url is https://github.com/willin/js.cool as far as I can see in willin's repositories..

willin:
oh, i'm sorry

README.md
```diff
@@ -1053,7 +1053,7 @@ Table of Contents
   * [freenom](https://freenom.com) — Free .tk,.ml,.ga,.cf and .gq domain.
   * [eu.org](https://nic.eu.org) — Free eu.org domain. Request is usually approved in 14 days.
   * [js.org](https://js.org) - Free js.org subdomains for GitHub Pages for the JavaScript community.
-  * [js.cool](https://github.com/willin/js-cool) - Free js.cool subdomains for GitHub Pages and Vercel.
+  * [js.cool](https://github.com/willin/js.cool) - Free js.cool subdomains for GitHub Pages and Vercel.
```
ripienaar, line 1056: It’s nice and all but I can’t really add it as this is just too restrictive. 


“ Your site must have a Chinese translation edition”

willin:
Your site ~~must~~(will be nice to) have a Chinese translation edition

README.md
ripienaar: Are these actually CNAMEs as stated? It doesn’t seem that way to me. It seems like a system of redirects rather than CNAME?

willin: - https://retrounhash.js.cool/
- https://leader.js.cool/
- https://xrkffgg.js.cool/
- and so on

ripienaar: Yes I saw and none of these are CNAMEs they resolve to the same IP.

ripienaar: Compare with for example js.org - that one makes actual CNAME dns entries. This one does not do  that.

ripienaar: To be clear: I am not listing a redirect based service.

willin: image


https://github.com/willin/js.cool/blob/master/main/active_cname.js

willin: perhaps because proxied?

ripienaar: Yes.

ripienaar: Others like js.org are not proxies

willin: same with js.org, it depends on users' willing: https://github.com/js-org/js.org/blob/master/cnames_active.js#L17-L22

ripienaar: Suggest you show this in similar comments in readme or here and I would need an example that works

willin: and it's almost the same implementation.

ripienaar:
Thank you

PR #2412, opened 2022-04-21
remove speeder cuz not available anymore

rexjohannes:
The Website is now an Ad Site so speeder.io is gone.


ripienaar:
Thanks :)

PR #2416, opened 2022-04-26
Added Livecycle to list of collaboration tools

zevi-r:
Great list! We added Livecycle which is an inclusive collaboration platform for CRX teams and open source maintainers/contributors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,7 @@ Table of Contents
   * [Keybase](https://keybase.io/) — Keybase is a cool FOSS alternative to Slack, it keeps everyone's chats and files safe, from families to communities to companies.
   * [Google Meet](https://meet.google.com/) — Use Google Meet for your business's online video meeting needs. Meet provides secure, easy-to-join online meetings.
   * [/meet for Slack](https://meetslack.com) - Start Google Meetings directly from Slack by using /meet in any channel, group or DM. Free without any limitations.
+  * [Livecycle](https://www.livecycle.io/) — Livecycle is an inclusive collaboration platform that makes workflows frictionless for cross-functional product teams and open source projects.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -352,6 +352,7 @@ Table of Contents
   * [Keybase](https://keybase.io/) — Keybase is a cool FOSS alternative to Slack, it keeps everyone's chats and files safe, from families to communities to companies.
   * [Google Meet](https://meet.google.com/) — Use Google Meet for your business's online video meeting needs. Meet provides secure, easy-to-join online meetings.
   * [/meet for Slack](https://meetslack.com) - Start Google Meetings directly from Slack by using /meet in any channel, group or DM. Free without any limitations.
+  * [Livecycle](https://www.livecycle.io/) — Livecycle is an inclusive collaboration platform that makes workflows frictionless for cross-functional product teams and open source projects.
```
ripienaar, line 355: this says 100% free while in beta but I cant see anything that says ther will be an ongoing free tier, can you provide a link pls?

zevi-r: We will always maintain a free tier. And for OSS project installations, it will remain free forever. What additional link would be helpful?

ripienaar: One that clearly states that on the site.  We need to be able to verify the validity of the list.

zevi-r: Here are a few links that explicitly state that the solution is free forever for OSS installations:
https://docs.livecycle.io/docs/open-source
https://docs.livecycle.io/docs/FAQs (see # 7 and 8 there)
https://oss.livecycle.io/
https://livecycle.io/blogs/oss/

ripienaar: Very nice. Thank you

PR #2420, opened 2022-04-30
Add Cloudflare to the email section

Sudobeast:
Cloudflare has [free email forwarding](https://blog.cloudflare.com/introducing-email-routing/).


ripienaar:
we prefer to list sites once only, so if their existing entry is not listing this lets amend that entry rather.  The list will be immense if we put them in many different places

Sudobeast:
understandable. is it good now?

ripienaar:
thank you

PR #2426, opened 2022-05-02
Remove jare.io

imba-tjd:
It became https://www.threecopies.com/

I also tried that, and the result is `400 Bad Request`


ripienaar:
thank you

PR #2434, opened 2022-05-10
Add Argonaut under PaaS category

guhans-argonaut:
Argonaut provides a forever free tier in their PaaS cloud deployment platform.
https://argonaut.dev/


ripienaar:
Thsnk you

imba-tjd:
I think it requires you to buy an IaaS by yourself.

ripienaar:
That doesn’t disqualify it - they offer a service to a problem many with IaaS would have. That service is delivered in the aaS manner so it’s suitable. There are quite a few like it on tbe list

PR #2448, opened 2022-05-18
Add uhostfull.com

raspiduino:
Add [uhostfull.com](https://www.uhostfull.com/) as a free web hosting service.

I personally used this in the past, and it's quite good. But it seems that a few people know about this site, so I will put it here.


FatGrizzly:
Sorry, We don't allow iFastNet resellers here.
image
Fake cPanel hosts aren't allowed on this list.

raspiduino:
I don't understand, it could be helpful for someone outside there.

raspiduino:
Ok I see. Thank you for widening my mind :)

PR #2454, opened 2022-05-27
Game Development Resources

vaimikpatel2908:


README.md
```diff
@@ -1616,6 +1616,19 @@ Table of Contents
 
 **[⬆ back to top](#table-of-contents)**
 
+## Game Development
```
ripienaar, line 1619: Should also add it to the top section pls.

Please keep in mind the scope of the list too, this might just be way out of scope


```
The scope of this particular list is limited to things that infrastructure developers (System Administrator, DevOps Practitioners, etc.) are likely to find useful. We love all the free services out there, but it would be good to keep it on topic. It's a bit of a grey line at times so this is a bit opinionated; do not be offended if I do not accept your contribution.```

vaimikpatel2908: No worries. I just wanted to add some resources which came into my sight.

vaimikpatel2908: @ripienaar Just added to the top section.

ripienaar:
thanks

PR #2457, opened 2022-05-29
Added Azure Free web app daily limit

ahmednrana:
Updated the limits as described here

[https://azure.microsoft.com/en-us/pricing/details/app-service/windows/](https://azure.microsoft.com/en-us/pricing/details/app-service/windows/)


ripienaar:
thank you

PR #2466, opened 2022-06-10
Add sslip.io DNS service

thomasdstewart:
Free DNS service that when queried with a hostname with an embedded IP address returns that IP address.


ripienaar:
nice, thank you

PR #2467, opened 2022-06-10
Added Fastlane CI/CD tool link

vaimikpatel2908:
Open Source CI/CD for android and ios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,6 +502,7 @@ Table of Contents
   * [codeship.com](https://codeship.com/) — 100 private builds/month, 5 private projects, unlimited for Open Source
   * [deployhq.com](https://www.deployhq.com/) — 1 project with 10 daily deployments (30 build minutes/month)
   * [drone](https://cloud.drone.io/) - Drone Cloud enables developers to run Continuous Delivery pipelines across multiple architectures - including x86 and Arm (both 32 bit and 64 bit) - all in one place
+  * [fastlane.tools](https://fastlane.tools/) - Fastlane is an open source platform aimed at simplifying Android and iOS deployment. Open Source CI/CD for development/release workflows. 
   * [LayerCI](https://layerci.com) — CI for full stack projects. 1 full stack preview environment with 5GB memory & 3 CPUs .
   * [ligurio/awesome-ci](https://github.com/ligurio/awesome-ci) — Comparison of Continuous Integration services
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -504,6 +504,7 @@ Table of Contents
   * [codeship.com](https://codeship.com/) — 100 private builds/month, 5 private projects, unlimited for Open Source
   * [deployhq.com](https://www.deployhq.com/) — 1 project with 10 daily deployments (30 build minutes/month)
   * [drone](https://cloud.drone.io/) - Drone Cloud enables developers to run Continuous Delivery pipelines across multiple architectures - including x86 and Arm (both 32 bit and 64 bit) - all in one place
+  * [fastlane.tools](https://fastlane.tools/) - Fastlane is an open source platform aimed at simplifying Android and iOS deployment. Open Source CI/CD for development/release workflows.
```
ripienaar, line 507: Is this offered as a service with a free tier?

vaimikpatel2908: Yet it is totally free. Just need to set up workflows as usually we have to do for all CI/CD.

ripienaar: It does not look like a SaaS to me

vaimikpatel2908: When we use Codemagic or other CI/CD for continuous integration and continuous deployment we need to setup a yaml which keeps config for CI/CD. so, here we also do have to setup AppFile and Fastfile for CI/CD. But other platforms have limites build minutes or projects. This is totally free for unlimited projects/minutes.

vaimikpatel2908: But it is just for android and iOS applications

vaimikpatel2908: If you have any doubt checkout the Codemagic CI/CD set up/installation and compare it with fatlane. You will get it. Codemagic has two options: 1. Using web GUI  2. Using yaml file. Fastlane just provides second way and for initialization they have fatlane cli which helps for initial setup.

vaimikpatel2908: For locally it works with just cli command and distributes app automatically. Bu if you want to automate then just need to configure github workflows. Check out following links.
https://docs.flutter.dev/deployment/cd
https://docs.fastlane.tools/

vaimikpatel2908: Definitely. I would not say that it is a standalone service but definitely good alternative for the paid services.

ripienaar: Please review the purpose of the list. It’s for SaaS.

PR #2473, opened 2022-06-15
Update README.md

someguy9:


someguy9:
Makes sense, want me to remove that one?

someguy9 (commented):
removed titlecap

PR #2478, opened 2022-06-20
Added `AO.ms` to `Dev Blogging Sites`

ao:


FatGrizzly:
The list is to list blogging sites (where users can start their own blogs, though I am unsure)

Let @ripienaar decide on this.

ripienaar:
Yes, I dont think its suited to this list, thanks for the contribution.

PR #2480, opened 2022-06-21
Add Poly Pizza

Chikanz:
Site that has a few thousand free CC-BY and CC0 models in obj/fbx/gltf formats. Useful for games, webgl etc


ripienaar:
thank you

PR #2488, opened 2022-06-28
Added elmah.io to Crash and Exception Handling tools

ThomasArdal:


ripienaar:
thanks

README.md
```diff
@@ -745,6 +745,7 @@ Table of Contents
 
   * [CatchJS.com](https://catchjs.com/) - JavaScript error tracking with screenshots and click trails. Free for open source projects.
   * [bugsnag.com](https://www.bugsnag.com/) — Free for up to 2,000 errors/month after the initial trial
+  * [elmah.io](https://elmah.io/) — Error logging and uptime monitoring for web developers. Free Small Business subscription for open source projects.
```
ripienaar, line 748: Looking again it seems to aply for this you have to go through a chat system? That's not really ok, most sites that has free for open source has a github onboarding where they get the license from the api.

If it's a "chat with us" signup, I am afraid I have to revert the merge

ThomasArdal: Correct. Some kind of proof that the user has an active open-source project. The chat function is not meant as a way to convince them to pay or anything like that. Everyone who fulfills the requirements gets a subscription. If this is not enough to be on the list, I understand why you want to remove the service. But currently, this is implemented to make it simple to get the subscription.

ripienaar: Yeah unfortunately we've had services with such sign ups and had complaints so removed them until they had an automated signup.

As detailed above its possible to get confirmation that a project is open source from the GH API, so this is something that can be automated.

PR #2492, opened 2022-07-05
Remove DDoS Guard

RelatedTitle:
It looks like they don't provide a free plan anymore. The lowest tier is their basic plan for $100 monthly.


ripienaar:
thank you

PR #2497, opened 2022-07-08
Update README.md

janlepsky:
I moved mogenius from Web Hosting to PaaS and edited the description to describe the platform more accurately.


ripienaar:
nice thanks

PR #2506, opened 2022-07-13
Add Rubocop

yaljazairi:
Add `Rubocop` as Code Quality. It's the best tool to follow Ruby style guide as well as suggesting a better way for implementing Ruby methods the way it should be done in Ruby.

Thanks for the efforts of managing this helpful document.


ripienaar:
This list is for SaaS offerings. Rubicon is not a SaaS offering.

PR #2512, opened 2022-07-17
Update PULL_REQUEST_TEMPLATE.md

FatGrizzly:
This should save us some time and also makes much easier reviewing PR's.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,6 +10,12 @@ would find useful.
 
 We do not anymore accept cPanel like PHP+MySQL hosting services.
 
+### New Service description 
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,6 +10,12 @@ would find useful.
 
 We do not anymore accept cPanel like PHP+MySQL hosting services.
 
+### New Service description 
+<!-- Please fill all these details, helps us review and merge your PR quickly -->
+Service name:
+Service website:
+Service pricing page:
```
ripienaar, line 17: Problem is once it’s too long no one will read it. I think my initial one was already too wrong

FatGrizzly: Fair. We probably need to trim it. I will try to trim it if I get some free time.

PR #2515, opened 2022-07-19
Added the cat-api and the dog-api

AbdAsh:

would find useful.

### New Submission Check List

### Thank You


ripienaar:
Thank you. I am not sure either of these have value within tbe ststed goal of the list

AbdAsh:
I thought they could be beneficial in some pre-tech-interview projects that are required when you're applying to some firms. So they could be quite a fun addition that could also show some character in your project. But in the end, it is up to you guys.

PR #2528, opened 2022-07-24
update readme to include Payload CMS

exponent42:

would find useful.

### New Submission Check List

 > It's currently self-hosted, but their examples and documentation make this really trivial.

They're also actively working on Payload Cloud which I expect will have a generous free tier, this will hopefully be out soon.

### Thank You


ripienaar:
This list is ONLY for SaaS services.

Thanks for the contribution but its not suitable for this list.

PR #2532, opened 2022-07-26
add upstash - qstash

chronark:
qStash is a new service that allows receiving events from in serverless functions, so you do not need a long running process to consume from a queue anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1124,7 +1124,7 @@ Table of Contents
    * [FaunaDB](https://fauna.com/) — Serverless cloud database, with native GraphQL, multi-model access and daily free tiers up to 100 MB
    * [HarperDb](https://harperdb.io/) — Serverless cloud database, with dynamic schema based on JSON, 3000 IOPS with 1GB storage
    * [heroku.com](https://www.heroku.com/) — PostgreSQL as a service, up to 10,000 rows and 20 connections free (provided as an "addon," but can be attached to an otherwise empty app and accessed externally)
-   * [Upstash](https://upstash.com/) — Serverless Redis with free tier up to 10,000 requests per day, 256MB max database size, and 20 concurrent connections
+   * [Upstash](https://upstash.com/) — Serverless Redis with free tier up to 10,000 requests per day, 256MB max database size
    * [MongoDB Atlas](https://www.mongodb.com/cloud/atlas) — free tier gives 512 MB
    * [redsmin.com](https://www.redsmin.com/) — Online real-time monitoring and administration service for Redis, Monitoring for 1 Redis instance free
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -645,6 +645,7 @@ Table of Contents
   * [pubnub.com](https://www.pubnub.com/) - Swift, Kotlin and React messaging at 1 million transactions each month. Transactions may contain multiple messages.
   * [eyeson API](https://developers.eyeson.team/) - A video communication API service based on WebRTC (SFU, MCU) to build video platforms. Allows RealTime Data Injection, Video Layouts, Recordings, a fully featured hosted web UI (quickstart) or packages for custom UIs. Has a [free tier for developers](https://apiservice.eyeson.com/api-pricing) with 1000 meeting minutes a month.
   * [Upstash Kafka](https://upstash.com/kafka) - Serverless Kafka Cloud offering with per-request-pricing. It has a free tier with max 10000 messages per day.
+  * [Upstash qStash](https://upstash.com/qstash) - Push based messaging and scheduling over HTTP for serverless and edge runtimes with 50 requests per day for free.
```
ripienaar, line 648: Maybe add upstash and then list all its free tiers rather than just one

chronark: yeah good idea, I pushed the changes

README.md
```diff
@@ -1122,7 +1124,7 @@ Table of Contents
    * [FaunaDB](https://fauna.com/) — Serverless cloud database, with native GraphQL, multi-model access and daily free tiers up to 100 MB
    * [HarperDb](https://harperdb.io/) — Serverless cloud database, with dynamic schema based on JSON, 3000 IOPS with 1GB storage
    * [heroku.com](https://www.heroku.com/) — PostgreSQL as a service, up to 10,000 rows and 20 connections free (provided as an "addon," but can be attached to an otherwise empty app and accessed externally)
-   * [Upstash](https://upstash.com/) — Serverless Redis with free tier up to 10,000 requests per day, 256MB max database size, and 20 concurrent connections
+   * [Upstash](https://upstash.com/) — Serverless Redis with free tier up to 10,000 requests per day, 256MB max database size
```
ripienaar, line 1127: we can remove this one now - we list services once only

chronark: That's a different product with a different free tier.

chronark: I see `pusher.com` also added under `BaasS` and `Messaging`, so I assumed it was fine to add different products under different categories.

ripienaar: We used to and are removing them as we come across them - the list is enormous already so we do not add more such multi listings.

ripienaar: Neither is great lets be honest, but lets pick which one you think is best.  It's a difficult situation but with the size of the list we gotta make some choices

PR #2534, opened 2022-07-27
Removed tebi.io from  IaaS

nextpier:

would find useful.

### New Submission Check List

### Thank You

---   
---  
tebi.io has now  a free tier + pay-as-you-go as basic option. It expires in 14 days.  After that period data becomes read-only, and requires card (no prepaid accepted)  to go on.  I think it should be removed from this list.

[https://tebi.io/#faq](https://tebi.io/#faq)  

tebi io-1


rachit1601:
But the free teir still hai 25GB storage and 250gb transfer

rachit1601:
https://tebi.io/#prices

ripienaar:
```In the Free Tier you have 25 GB of storage (with 2 copies) and 250 GB of traffic. This volume of services is credited every month for free. If you are using more storage or bandwidth, you just need to pay for the amount that exceeds the Free Tier.
```

Still seems legit to me and being that you pay for overages seems fine to ask for CC, other services do this too.

nextpier:
Ok :-)
Maybe just mention the mandatory card

PR #2537, opened 2022-07-31
Email: Smtpjs.com

OgbeniHMMD:


ripienaar:
What is the SaaS service here? Looks like just a JS library.

OgbeniHMMD:
"It is not possible to send emails using SMTP servers with JS or Jquery. SMTPJs gives us API but eventually we are using their servers to hit the SMTP. actually we are sending data to their server which further sends data to SMTP servers.”

Ref: https://stackoverflow.com/questions/46936506/smtpjs-api-not-working-is-there-any-way-to-send-emails-using-smtp-server-with-j

ripienaar:
I am aware how SMTP works. In the scenario if you offered a free SMTP server that would be a SaaS. A JS library, unless a fromtend to a SaaS with a free tier, is not a SaaS

OgbeniHMMD:
Oh! I think I get your point now.

ripienaar:
Thank you

PR #2542, opened 2022-08-03
Remove Arvan Cloud from CDN and Protection

rayanlancelotap:
This Company is based in Iran and has a bad reputation for working with the regime. It has an excellent part in making the national internet and filtering. it is getting government funds. In Iran, every company should operate under the regime's rules. please don't advertise these services. (I'm Iranian by the way)

### New Submission Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -919,11 +919,6 @@ Table of Contents
 
 ## CDN and Protection
 
-  * [Arvan Cloud](https://arvancloud.com/) — Offers cloud related services (CDN,Cloud DNS, PaaS, Security etc.). Free plan offers :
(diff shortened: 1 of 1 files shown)


ripienaar:
Thanks was not aware of their background but saw now it's based there, indeed not happy having it on the list.

DreamOfIce:
I have also used their CDN. To be fair, the speed is really good.
I was a little surprised to learn about their cooperation with the Iranian government.
However, it seems that it is also a gold sponsor of [jsdelivr](https://www.jsdelivr.com/sponsors)?

PR #2554, opened 2022-08-11
Update humio logging limit per day

akonyer:
Changing the 2GB to 16GB that is now included with the free account.


ripienaar:
Thank you

PR #2557, opened 2022-08-16
Removed Katacoda

1Prototype1:
Katacoda.com is now closed :(


ripienaar:
such a shame, was a very good tool.  thanks for the update

PR #2560, opened 2022-08-19
Removed Yandex.Connect

rufanov:
During 2021, Yandex.Connect was reorganised to two separate platforms - Yandex.360 and Yandex.Cloud. 

Mentioned "Free email and DNS hosting for up to 1,000 users" was moved to Yandex.360, but in December 2021 Yandex dropped free service, so it's not free anymore for new users - https://360.yandex.com/business/tariff/


ripienaar:
thank you

PR #2568, opened 2022-08-25
Remove heroku because it is no more free

PeterPorzuczek:
Heroku will not be free anymore, check Here: https://blog.heroku.com/next-chapter


sharshi:
That was fast ☹️

PeterPorzuczek:
@sharshi I had to see what alternatives there were and came across this repo then I thought I'd do an update right away

FatGrizzly:
thank you

PR #2570, opened 2022-08-26
Add domcloud.co to list of free PaaS services

willnode:
### New Submission Check List


willnode:
Oh, I see `Web Hosting` category too. What's the difference? Should I move it there too?

ripienaar:
We list each entry once only so if you want to move it go ahead else we can merge as is

willnode:
I think it gonna be closer to PaaS sooner or later so I think we're good to merge as is 👍

ripienaar:
Thank you

PR #2571, opened 2022-08-27
bunny.net free service

raestrada95:


raestrada95:
Hi, the bunny dns pricing https://bunny.net/pricing/dns/

ripienaar:
We tend to list companies that offer multiple services once only in the most appropriate place. Keeps the list manageable but maybe a bit harder to find things

PR #2583, opened 2022-08-30
Add privado as privacy management tool

hiteshbedre:
### New Submission Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -1612,6 +1612,7 @@ Table of Contents
   * [Iubenda](https://www.iubenda.com/) - Privacy and cookie policies along with consent management. Free tier offers limited privacy and cookie policy as well as cookie banners.
   * [Cookiefirst](https://cookiefirst.com/) - Cookie banners, auditing, and multi-language consent management solution. Free tier offers a one-time scan and a single banner.
   * [Ketch](https://www.ketch.com/) - Consent management and privacy framework tool. Free tier offers most features with a limited visitor count.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1612,6 +1612,7 @@ Table of Contents
   * [Iubenda](https://www.iubenda.com/) - Privacy and cookie policies along with consent management. Free tier offers limited privacy and cookie policy as well as cookie banners.
   * [Cookiefirst](https://cookiefirst.com/) - Cookie banners, auditing, and multi-language consent management solution. Free tier offers a one-time scan and a single banner.
   * [Ketch](https://www.ketch.com/) - Consent management and privacy framework tool. Free tier offers most features with a limited visitor count.
+  * [Privado](https://www.privado.ai/) - Tool identifies personal data as well as tracks the journey of processed data from the point of collection to defined sinks such as Third Parties, Databases, Logs, and APIs - thereby intelligently generating dataflows for your repository.
```
ripienaar, line 1615: Hard to see exactly but in the OSS version you run it locally/in CI and the output is JSON.  So is it really a SaaS? It also only support Java in the free version - worth mentioning assuming it is indeed a SaaS

hiteshbedre: privado tool has a different product range. Their data mapping tool comes under saas offering and cli tool is a core component behind saas offering. Also, their saas offering supports more than 20 languages.

hiteshbedre: Screenshot 2022-08-31 at 11 14 04 PM

The product range screenshot is attached here.

ripienaar: Yes. But what is the free feature here exactly. To me it looks like the CLI mainly?

hiteshbedre: yes, CLI is the free one.

ripienaar: If it doesn’t interact with a SaaS then it’s not suited to the list - and from the description I think it doesn’t.

PR #2585, opened 2022-08-31
Add other Free Resources

Licoy:
[Utils.fun](https://utils.fun/en) — All offline daily and development tools based on the browser's computing power, including watermark generation, screen recording, encoding and decoding, encryption and decryption and code formatting, are completely free and do not upload any data to the cloud for processing.


ripienaar:
thanks

PR #2609, opened 2022-09-18
Update WayScript free runtime hours

samgozman:
According to their price page https://www.wayscript.com/pricing they reduced free runtime hours from 744 to 100


ripienaar:
Thank you

PR #2610, opened 2022-09-18
form-data.com

danbars:
### New Submission Check List


README.md
```diff
@@ -884,6 +884,7 @@ Table of Contents
 **[⬆ back to top](#table-of-contents)**
 
 ## Forms
+  * [Form-Data](https://form-data.com) - No-code forms backend. Spam filter, email notification and auto-respond, webhooks, zapier, redirects, AJAX or POST and more. Free plan offers unlimited forms and 2000 submissions.
```
ripienaar, line 887: No, it offers 20 submissions per month without advertising the service.

danbars: I think you are looking at it the other way around 😉
Many services in this list give you a free tier with "branding", and then the paid tier allows you to "remove branding".
Just as an example, look at HeyForm's pricing https://heyform.net/pricing
You'll see "remove branding" in the first tier. And this is not an unusual thing.
Most if not all services in the forms section, send branded emails, and many don't even mention this. 
So another way to look at it, is that my service actually gives you 20 non-branded submissions in addition to 2000 branded ones. This is an extra mile that I'm going compared to others.

ripienaar: That’s fine, you can close the PR of you don’t want to put a correct pricing description. 

But I am not here to argue about it.

danbars: Ok, I've refined the description. Let me know if that's ok. 
Thanks.

ripienaar:
thank you

PR #2615, opened 2022-09-20
Remove Jotcode since it's shutdown

DriftMark:
### New Submission Check List

 - [ ] The service has contact details of those running it and a privacy policy


ripienaar:
Thank you

PR #2621, opened 2022-09-28
Remove duplicated entry

bitsnaps:
ably appears twice, once at "BaaS", another at "Messaging and Streaming", I think the later is more appropriate.


ripienaar:
Great thank you

PR #2623, opened 2022-09-30
Update README.md

Heccubernny:
## Added:
  - readme.so
  - swagger inspector
  - JsonCrack

### New Submission Check List


ripienaar:
Thanks for this, unfortunately you've changed every line in the file and while most of the formatting cleanup seems good somehow now every entry under `APIs, Data and ML` is rendered with a blank line between etc.

Maybe best just to add the entries and then in a followup commit do the markdown fixes so we can try to contain things a bit

Heccubernny:
I' sorry. I think it is the prettify extension. Can I rework on it and send another pull request. Thank you

ripienaar:
Yeah of course

PR #2643, opened 2022-10-10
Updated Okta offerings

hkbertoson:
Updated Okta Offerings from 1000 monthly active users to 7000


FatGrizzly:
Their website says this, You'll be redirected to Auth0.com, which is becoming the Okta Customer Identity Cloud.

hkbertoson:
> Their website says this, You'll be redirected to Auth0.com, which is becoming the Okta Customer Identity Cloud.

Yes. That is correct. But they still have the free 7000 monthly active users.

FatGrizzly:
> > Their website says this, You'll be redirected to Auth0.com, which is becoming the Okta Customer Identity Cloud.
> 
> Yes. That is correct. But they still have the free 7000 monthly active users.

I feel like we can remove okta from the list, since the website just redirects to auth0

PR #2650, opened 2022-10-16
PaaS: add Replit

89z:
### New Submission Check List


FatGrizzly:
replit is already listed under ide

89z:
Its both, so I think it would be helpful to be listed as both

FatGrizzly:
Unfortunately, We don't list services twice

PR #2659, opened 2022-10-19
Add https://game-icons.net

pyko:
### New Submission Check List


pyko:
This is a resource, not a service (so SaaS vs self-hosted not relevant). Similar to some other links which are good resources.

FatGrizzly (approved):

PR #2680, opened 2022-11-04
Update README.md about the changed free limit of pdfmonkey

Thenafi:
The pdfmonkey free tier limit is 300 per month now. 
https://is.gd/jjuPoO here is a screenshot of the updated price


ripienaar:
thanks

PR #2683, opened 2022-11-14
Remove asayer.io

ruggi99:
Remove asayer.io because it moved its domain to `openreplay.com`.
This site does not offer (anymore I think) free tier but only 14 days free trial.


ripienaar:
thank you

PR #2692, opened 2022-11-21
Added QR Code Robot - free QR Code generator

beniaminp:
### New Submission Check List

 - [ ] The service has contact details of those running it and a privacy policy


imba-tjd:
I don't think this is useful. QR code is just a way to encode. Unless you actually implement `file sharing (pdf, mp3, video)`, it's not considered as a service.

ripienaar:
We have a few services like this here already - and the author states he wants to do things like open tracking (I guess he means click tracking) etc in future which is something I hadnt seen.

Agree its on the margins of usefulness as a SaaS though.

PR #2711, opened 2022-12-06
Add pre-commit.ci

zer0-x:
### New Submission Check List


ripienaar:
This seems super limited being just focussed on one framework? 

I don’t really know this framework so just asking if that’s the case? Then I don’t think it’s sensible on a generic list like this

zer0-x:
> This seems super limited being just focussed on one framework?

It's a framework that let you easily do the work of git hooks in your git repo so you can run linters and code formaters and other things to identify simple issues before submission to code review. It point out issues in code such as missing semicolons, trailing whitespace, debug statements, security things, and unneeded variables... (Depending on what linters you configured it to use)
This allows a code reviewer to focus on the architecture of a change while not wasting time with trivial style nitpicks.

You can configure it to use any linter to run a check for every git commit or pull-request.

It's not limited, you have a lot of linter to run using it to check for code in a lot of programming languages.

`pre-commit.ci` will run the linters that you configured for every pull-request and commit and it will tell you if any linter got an update.

ripienaar:
Yes, but rather than a generic thing like a CI service, its a CI integration for one specific framework only. 

I think it shows clearly here `...nothing is needed beyond the .pre-commit-config.yaml file you already have`, no, very few people have that already, too few to be considered for a generic list like this imo.

I might be wrong, but I think I'll pass on this one thanks for the contribution though

PR #2722, opened 2022-12-11
Add forwarddomain.net

willnode:
### New Submission Check List

 - [X] The service has contact details of those running it and a privacy policy


README.md
```diff
@@ -1072,6 +1072,7 @@ Table of Contents
   * [dnspod.com](https://www.dnspod.com/) — Free DNS hosting.
   * [duckdns.org](https://www.duckdns.org/) — Free DDNS with up to 5 domains on the free tier. With configuration guides for various setups.
   * [dynu.com](https://www.dynu.com/) — Free dynamic DNS service
+  * [forwarddomain.net](https://forwarddomain.net/) — Free redirect domain service using only DNS records
```
ripienaar, line 1075: "using DNS only with 301 HTTP Redirects" doesnt really make sense.

Anyway, we've had a few of these apply here and we rejected them as not being on topic really, so will do this one as well, sorry.

willnode: Ok, I understand. Thanks 👍

PR #2733, opened 2022-12-17
Remove Atom

ItsNiceCraft:
Atom was killed by GitHub on December 15, 2022

https://github.com/atom/atom/releases/tag/v1.63.1
https://github.blog/2022-06-08-sunsetting-atom


ripienaar:
thank you

PR #2745, opened 2022-12-22
Remove gitea.com

1toldyou:
> This Gitea instance primarily serves the development needs of the Gitea project itself

> While we have open sign-ups, this server is not an equivalent to gitlab.com or github.com.


ripienaar:
thank you

PR #2753, opened 2022-12-28
update

ayushch80:
you might remove `livesession.io` as they have moved to 14 days free trial
also `hotjar.com` have limited requests from 5000/day to 1050/month


ripienaar:
ok, in that case go for it, remove livesession

ayushch80:
removed livesession.io

ripienaar:
thanks!

PR #2765, opened 2023-01-10
Add MessengerX.io as a resource

unixguru2k:
Add MessengerX.io as a resource under the APIs, Data, and ML section.  This API is a free resource for developers or creators interested in build, growing and personalizing AI powered characters or chat apps.

## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+ 
   * [JSON IP](https://getjsonip.com) — Returns the Public IP address of the client it is requested from. No registration required for free tier. Using CORS data can be requested using client side JS directly from browser. Useful for services monitoring change in client and server IPs. Unlimited Requests.
   * [konghq.com](https://konghq.com/) — API Marketplace and powerful tools for private and public APIs. With the free tier, some features are limited such as monitoring, alerting and support.
   * [Kreya](https://kreya.app) — Free gRPC GUI client to call and test gRPC APIs. Can import gRPC APIs via server reflection.
+    * [MACHAAO aka MessengerX.io](https://github.com/machaao/) - Build, customize and grow deeply personalized conversational characters / Chat Apps with our free to use cloud platform APIs for developers / creators.
   * [Lightly](https://www.lightly.ai/) — Improve your machine learning models by using the right data. Use datasets of up to 1'000 samples for free.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -262,6 +262,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+
   * [JSON IP](https://getjsonip.com) — Returns the Public IP address of the client it is requested from. No registration required for free tier. Using CORS data can be requested using client side JS directly from browser. Useful for services monitoring change in client and server IPs. Unlimited Requests.
   * [konghq.com](https://konghq.com/) — API Marketplace and powerful tools for private and public APIs. With the free tier, some features are limited such as monitoring, alerting and support.
   * [Kreya](https://kreya.app) — Free gRPC GUI client to call and test gRPC APIs. Can import gRPC APIs via server reflection.
+    * [MACHAAO aka MessengerX.io](https://github.com/machaao/) - Build, customize and grow deeply personalized conversational characters / Chat Apps with our free to use cloud platform APIs for developers / creators.
```
ripienaar, line 265: from what I can tell this requires a user to host a server - the platform calls out to the users server in order to make the bot work.

Right? So without such a server you cant make it work for free?

I dont see any pricing information here so I dont really know how to confirm this is free or not.

unixguru2k: Yes, you can sign up and create a character with or without the server.
And it's free for developers looking to build low code characters for fiction, history or learning in general.

ripienaar: OK, so I need to see some document that describes that clearly.

ripienaar: This is all good and well, but you'll appreciate that this list also need to be maintained.  And other users from time to time goes through these, visit some of the sites and later come back to edit the list should it no longer be valid information.

Now with this information being somewhat hidden in a way that you can only show it to me with screenshots and after having signed up, its just essentially unverifiable.

Once you have a clear page up somewhere that details the service and what is free we can consider adding it here, till then its just too unclear.

unixguru2k: Hi @ripienaar 
I do appreciate you maintaining this list.
Any recommendations or suggestions on what would you define in terms of a public listing?
Coz we don't ask credit cards, its a free sign up?
There is no hidden agenda here, we just need to know our api doesn't get abused hence there is a sign up
But i thought we were on the same page as far as helping developers with tools that supplement their knowledge?

ripienaar: We are on the same page that dev tools are great, even better when free.

I still need to be able to easily confirm the claim that yours is free without signing up.

So you'll have to make some kind of feature list and state it's free or something.

PR #2772, opened 2023-01-13
Remove hostman

handgun50:
Didn't found the free tier pricing

## Requirements


ripienaar:
Thank you seems like a free trial now

PR #2777, opened 2023-01-17
Remove awesomeness

WilliamDavidHarrison:
## Requirements

This should be removed as the website doesn't exist any more.


ripienaar:
Thanks

PR #2788, opened 2023-01-24
Bagisto added

Sanjaybhattwebkul:
>Bagisto added


FatGrizzly:
Please read the code of conduct while you open the PR.
We don't accept opensource repositories.

ripienaar:
To be clear, we do support open source, as long as the company behind it offer a SaaS hosted version of that open source and if that SaaS hosted version has a free tier.

Being able to self-host it is gravy, but its not enough to make it onto this list.

PR #2793, opened 2023-01-27
add pagerly

Falit:
## Requirements

 * [ yes ] This is Software as a Service not self hosted
 * [ yes ] It has a free tier not just a free trial
 * [ yes ] The submission mentions what is free
 * [ yes ] The submission is not already present in the list
 * [ yes ] The service has contact details of those running it and a privacy policy


ripienaar:
thank you

PR #2815, opened 2023-02-03
Update Koyeb

hashlash:
Koyeb's currently requires credit card information to signup.

Quote from the CEO's [tweet](https://twitter.com/yann_eu/status/1612826949130371073):

> We require a credit card because it's a pay-per-use plan and to reduce abuse.


ripienaar:
Thanks, not that uncommon and seems like a good way to safe guard free tiers.

hashlash:
@ripienaar 
Actually, I was inspired by the change from #2303.

Btw, is there any plan to make the list more comparable? Like listing the feature of each category, or perhaps put the list and details of what the free tier got as a table.

ripienaar:
Improving the list is always on the cards but we have to balance it with the cost of maintenance, what you propose sounds like a lot of work and I doubt it will be kept up to date.  It would be really great though I agree

PR #2817, opened 2023-02-08
Add Nango to the Auth section

rguldener:
## Requirements


rguldener:
Direct link to the service's website: https://www.nango.dev
Pricing page: https://www.nango.dev/pricing

FatGrizzly:
sounds like a self hosted option only? cloud one seems paid.

ripienaar:
Seems to me cloud includes 10 for free connections then $0.5/month. But if it means like 1 single auth request is a "connection" then lol, no thats just not a free service.

ripienaar:
ahh, `1 A connection is an active OAuth token, where Nango handles the OAuth flow for the end-user, secure token storage and automatic refreshing.` so 10 active oauth tokens a monht, I dont think thats enough to be useful to anyone.

PR #2822, opened 2023-02-10
Remove unscan.co

1toldyou:
Domain expired, their [npm package](https://www.npmjs.com/package/unscan) marked as deprecated.


FatGrizzly (approved):
thanks

PR #2824, opened 2023-02-11
Add OpenAPI.security

glimow:
## Requirements


ripienaar:
Thank you

PR #2839, opened 2023-02-23
https://www.softacheck.com is DNS_PROBE_FINISHED_NXDOMAIN,

chirag127:
changing the link to the GitHub app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,7 +462,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+ 
 
 ## Code Quality
 
-  * [SoftaCheck](https://www.softacheck.com/) — An online tool that performs static analysis for C/C++ code using open source tools such as cppcheck and clang-tidy and automatically generates code documentation for users using doxygen. This tool is free for use.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -462,7 +462,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+
 
 ## Code Quality
 
-  * [SoftaCheck](https://www.softacheck.com/) — An online tool that performs static analysis for C/C++ code using open source tools such as cppcheck and clang-tidy and automatically generates code documentation for users using doxygen. This tool is free for use.
+  * [SoftaCheck](https://github.com/apps/softacheck) — An online tool that performs static analysis for C/C++ code using open source tools such as cppcheck and clang-tidy and automatically generates code documentation for users using doxygen. This tool is free for use.
```
ripienaar, line 465: think we can just remove it at this point tbh.

chirag127: But the tool seems to be working.

ripienaar: if they dont love it enough to fix their stuff, neither do i :)

chirag127: okay, I will the edit the pr with removal

chirag127: I also thought I should remove the tool before making the pr but then thought maybe you will say that the tool is working on the github , so, we should not remove.

ripienaar:
Thank you :)

PR #2559, opened 2022-08-18
Update jsonbin.io

handgun50:
The old description didn't match with current pricing. Not totally free for lifetime. Only free for the first 10.000 requests. After that pay. Don't know it is still considered free or not

### New Submission Check List

Please ensure your submission ticks all of these boxes:

 - [ ] This is Software as a Service not self hosted
 - [ ] It has a free tier not just a free trial
 - [ ] The submission mentions what is free
 - [ ] The submission is not already present in the list

PR #2870, opened 2023-03-16
Added jsonbin.io

fiorins:
add jsonbin.io

## Requirements


ripienaar:
Maybe it’s because I am on mobile but I can’t see any details of what is free on their site.

fiorins:
These are its plans:

Screenshot 2023-03-16 at 8 59 51 AM

ripienaar:
So you get 10000 requests free and that’s the end of the free plan? That’s how it reads to me in the FAQ. 

That won’t qualify for the list. That’s basically a free trial.

handgun50:
actually jsonbin already removed on #2559 because it only gives 10.000 request only

PR #2886, opened 2023-03-29
Fixed ZeroTier number of clients

masoncfrancis:
Updating existing listing: ZeroTier

ZeroTier has reduced their number of free clients from 100 to 25.


ripienaar:
thank you

PR #2902, opened 2023-04-12
Rename LogDNA to Mezmo

ap-1:
[LogDNA is now Mezmo](https://www.mezmo.com/logdna)


handgun50:
I think we should remove it because no retention is not quite useful. What do you think?

ripienaar:
Yeah good call, I also just looked at their site again and its very opaque and useless, no pricing information and only talks about trials.  

Lets remove it, want to send a PR @handgun50 ?

PR #2905, opened 2023-04-13
Add Raivo OTP

ap-1:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,6 +649,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+ 
   * [Authgear](https://www.authgear.com) - Bring Passwordless, OTPs, 2FA, SSO to your apps in minutes. All Front-end included. Free up to 5000 MAUs.
   * [Authress](https://authress.io/) — Authentication login and access control, unlimited identity providers for any project. Facebook, Google, Twitter and more. First 1000 API calls are free.
   * [Authy](https://authy.com) - Two-factor authentication (2FA) on multiple devices, with backups. Drop-in replacement for Google Authenticator. Free for up to 100 successful authentications.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -649,6 +649,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+
   * [Authgear](https://www.authgear.com) - Bring Passwordless, OTPs, 2FA, SSO to your apps in minutes. All Front-end included. Free up to 5000 MAUs.
   * [Authress](https://authress.io/) — Authentication login and access control, unlimited identity providers for any project. Facebook, Google, Twitter and more. First 1000 API calls are free.
   * [Authy](https://authy.com) - Two-factor authentication (2FA) on multiple devices, with backups. Drop-in replacement for Google Authenticator. Free for up to 100 successful authentications.
+  * [Raivo OTP](https://raivo-otp.com/) — Completely free and open source 2FA authenticator with rich Apple integration. Drop-in replacement for Google Authenticator on iOS.
```
ripienaar, line 652: This is cool and all but its like essentially self hosted software there's no SaaS behind it like say Authy just above it.  So I am not sure it would qualify.

ap-1: I just figured some users would opt for an open source alternative, given that it provides the same service Authy does. 

In addition, the description for Authy mentions it's only free for 100 authentications --- for someone like me, who uses 2FA for everything, their free tier would hardly last a few weeks (sounds more like a free trial, compared to Raivo which is always free). After it expires, Authy [doesn't make it easy to export your keys](https://github.com/pluja/awesome-privacy/), which sounds like vendor lock-in to me.

ripienaar: Yeah I get it, but imagine we added every standalone open source app thats kind of tangantly related to this list? We'd have 10s of thousands of entries and no1 will find anything.

So this list is focussed to SaaS (with a small exception for a few like IDEs), I am sure there are other lists this would be very welcome on

https://github.com/sindresorhus/awesome

ap-1: I see, I'll take a look at that list

PR #2912, opened 2023-04-15
Move patr.cloud and add details on anvil.works

arshadrr:
Patr.cloud's pricing page (https://patr.cloud/pricing ) only offers 3 static sites for free. Clicking on deployments you either connect a cloud service like AWS or use their service starting at $5. No longer free.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1003,7 +1003,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+ 
 
 ## PaaS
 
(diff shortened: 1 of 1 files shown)


ripienaar:
thank you

handgun50:
Actually patr is still offering free deployment. I'm still using it. $5 shows up when deployment because we add a disk. If you didn't add a disk, it will show $0

PR #2914, opened 2023-04-17
Updated Scalr description and category.

DicsyDel:
## Requirements


ripienaar:
Thanks

PR #2919, opened 2023-04-20
Remove EVA (eva.pingutil.com)

ItsNiceCraft:
Removed EVA (eva.pingutil.com) because it has been down for several months now. pingutil.com is also down.


ripienaar:
thank you

FatGrizzly:
@ripienaar the api still seems to work from my end, with an invalid cert

ripienaar:
hmm, now that I look again (I checked only pingutil.com) I see eva is up for me. but with the main site down, its just a below par service so inlined to remove it

ItsNiceCraft:
Weird, it's up again! Altough I checked a few months ago and it was down as well, then, an hour ago I checked again and submitted this pr

PR #2925, opened 2023-04-28
Add `nudge-hook.net`

jimpil:
## Requirements

Note that this is a completely free & unlimited api (at least with respect to how many schedules one can create). As such, there is no signup required, no login screen, or api-key etc. In other words, the only piece of user-related info the server needs to know is simply an email address (of the job-owner) - no card details, or login credentials are stored anywhere!


jimpil:
By the way, the entry for `posthook.io` is outdated. There is no loger a free tier - it starts at $19. I'm not sure how you want to handle this though, as it no longer qualifies for this list.

ripienaar:
You can update yours to just remove it or send a new one.

jimpil:
Done

ripienaar:
thank you

PR #2954, opened 2023-05-23
Re-added HelioHost since free signups are available again

Kryd0s:
Added HelioHost under Web Hosting category.

## Requirements


ripienaar:
Nope, not adding it. Had enough spammers and general abuse here from it or people who support it.

ABUCKY0:
@FatGrizzly

ABUCKY0:
> Nope, not adding it. Had enough spammers and general abuse here from it or people who support it.

Can you elaborate on the spammers and general abuse from HelioHost supporters? I was only able to find one other mention of it in the pull requests on an unrelated PR comment.

ripienaar:
Yes, that one. Though I believe some PRs are not visible due to abuse claims maybe?

It also seems severely resource constraint etc.

Anyway, I am not adding it again.

Kryd0s:
One of the other web hosting listings you have only offers 50 MB storage, and HelioHost has 1000 MB storage. I don't understand how that is "severely resource constraint" in comparison.

Also ChippyTech, if that's the spammer you're referring to, has absolutely no affiliation with HelioHost. We have about 5k accounts right now, and we can't be held responsible for the behavior of every single one of them.

PR #2961, opened 2023-05-29
Update README.md

aloncarmel:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+ 
 * [ERD Lab](https://www.erdlab.io) —  Free cloud based entity relationship diagram (ERD) tool made for developers.
 * [What The Diff](https://whatthediff.ai) - AI-powered code review assistant. The free plan has a limit of 25,000 tokens per month (~10 PRs).
 * [Zipcodestack](https://zipcodestack.com) - Free Zip Code API and Postal Code Validation. 10000 free requests/month.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -349,6 +349,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+
 * [ERD Lab](https://www.erdlab.io) —  Free cloud based entity relationship diagram (ERD) tool made for developers.
 * [What The Diff](https://whatthediff.ai) - AI-powered code review assistant. The free plan has a limit of 25,000 tokens per month (~10 PRs).
 * [Zipcodestack](https://zipcodestack.com) - Free Zip Code API and Postal Code Validation. 10000 free requests/month.
+* [Wilco Sierra](https://trywilco.com/sierra) - A platform that generates engineering challenges for software engineers using AI.
```
ripienaar, line 352: Lots of "Try Siera" and "Start Free" etc on that page, but I cant find anywhere details of what happens after "start"? Like what do you get for free, when do you start paying etc.

Need the pricing/offering information to be clear.

aloncarmel: You dont. It's free there is no pricing or any plans to.

ripienaar: Suggest you get the messaging right on the page then "Start for free" etc suggests its not free free.  Happy to list it just need the terms clear so we can evaluate it correctly today and later

aloncarmel: The mistake is that the top right button start for free is for the platform in general. Sierra as the listing is totally free. Thanks for the comment, I will adjust.

ripienaar: cool, anything you can do to clarify its free would be great.

PR #2978, opened 2023-06-10
added cloudflare R2

blank1729:
I still think it's in here somewhere, but I couldn't find it

## Requirements

 * [ x ] This is Software as a Service not self hosted
 * [ x ] It has a free tier not just a free trial
 * [ x ] The submission mentions what is free
 * [ x ] The submission is not already present in the list
 * [ x ] The service has contact details of those running it and a privacy policy

### DIFF
--- a/README.md
+++ b/README.md
@@ -1151,6 +1151,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+ 
 
 ## IaaS
 
+  * [Cloudflare R2](https://www.cloudflare.com/products/r2/) - Free 10GB Object storage with S3-compatible API and Zero egress fee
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1151,6 +1151,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+
 
 ## IaaS
 
+  * [Cloudflare R2](https://www.cloudflare.com/products/r2/) - Free 10GB Object storage with S3-compatible API and Zero egress fee
```
ripienaar, line 1154: Better to add it to the existing cloudflare entry.  Thanks!

blank1729: The existing cloudflare entry is under CDN and ... 
I don't think it fits there
If I were to search for object storage, I won't be searching it in CDN

ripienaar: At the top is a whole dedicate CF entry listing all their stuff. We try to list same company ilonly once.

PR #2987, opened 2023-06-26
DNS - Add Dual.my

DennisKidman:
## Requirements

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -30,8 +30,8 @@
 <!-- This is only for new submissions -->
 <!-- Please ensure your submission ticks all of the requirements -->
 
- * [ ] This is Software as a Service not self hosted
- * [ ] It has a free tier not just a free trial
- * [ ] The submission mentions what is free
- * [ ] The submission is not already present in the list
- * [ ] The service has contact details of those running it and a privacy policy
+ * [X] This is Software as a Service not self hosted
+ * [X] It has a free tier not just a free trial
+ * [X] The submission mentions what is free
+ * [X] The submission is not already present in the list
+ * [X] The service has contact details of those running it and a privacy policy
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -30,8 +30,8 @@
 <!-- This is only for new submissions -->
 <!-- Please ensure your submission ticks all of the requirements -->
 
- * [X] This is Software as a Service not self hosted
- * [X] It has a free tier not just a free trial
- * [X] The submission mentions what is free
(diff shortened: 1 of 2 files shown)


ripienaar:
thank you

DennisKidman:
Hello,

I understand your concerns. Please note that the website is still in its early stages which is why you're currently seeing stats for only one domain. The goal is to promote it on the free-for-dev list to attract new users and continually improve the site based on user feedback and interaction.

Regarding your second point, our setup consists of three DNS servers, with one acting as the master and two as slave servers. We have decided to use Cloudflare primarily to mask our IP. As this is our first live project, we want to minimize our exposure to potential DDOS attacks. At present, only one DNS server is public without a Cloudflare proxy, which might change for the other servers as we continue to test and optimize.

I'm not entirely sure about your reference to "Cloudflare frontends"; could you please elaborate on that?

ripienaar:
You get a kind of DNS host that just relies on cloud flare dns service APi and often hosted on their free tier which is very limited. We don’t support that. 

Still questions of why all the NS records when they point at one place. 

For DNS services I am not really too happy to add such early services here as it’s such a fundamental component I want to list known to be reliable services here

DennisKidman:
> You get a kind of DNS host that just relies on cloud flare dns service APi and often hosted on their free tier which is very limited. We don’t support that.

I get where you're coming from with the concerns about relying on Cloudflare's DNS service API, especially if it's on their free tier. We're doing things a bit differently though - we're self-hosting our DNS server on a Tier 1 host in Europe. Cloudflare for us is just a protective shield to hide our IP. We've also got one of our slave servers out in the public domain, without Cloudflare's proxy.

> Still questions of why all the NS records when they point at one place.

regarding your query about all NS records pointing to one place it's part of our strategy to obscure our NS IP. The CNAME record we utilize holds both A and AAAA for our DNS server, serving as a cloak for our NS IP. As we get more comfortable and confident with the system, we'll consider making more servers publicly accessible.

> For DNS services I am not really too happy to add such early services here as it’s such a fundamental component I want to list known to be reliable services here

Absolutely, I get your concern about introducing new services to the list. The fundamental nature of DNS and the necessity for reliability is not lost on me. I want you to know that I've poured countless hours over many months into this project. I'm deeply invested and committed to providing this service free of cost, without any ads.

If youre still unsure, I totally respect that and Im more than willing to wait another month before making another request. But if you're up for it, maybe you could try adding some records and using our API first? That way, you'll get to see the level of effort and dedication that has gone into building this project. Either way, I appreciate you taking the time to consider at our PR.

ripienaar:
I would prefer you come back a month later tbh. 

All this obscuring your details (it’s trivial to find real details (as I showed you) I am not happy with that. And it’s useless as DDOS protection as IPs are still resolvable else nothing would work.

PR #2998, opened 2023-07-02
Update doprax

jneeee:
When create an app, a "app space" need to be created.
Got a remind when create an app space:

image


FatGrizzly (approved):

FatGrizzly:
thanks

PR #3011, opened 2023-07-26
Remove Okteto cloud

jaens:
No free tier, only has a trial now, for business e-mails only, see:

https://www.okteto.com/free-trial/
https://www.okteto.com/pricing/


ripienaar:
Thank you

PR #3020, opened 2023-08-07
Remove Ns1 from the Free DNS list

madhuryadutta:
Remove the NS1 DNS service. According to ns1.com ,currently the new free account (Developer Plan) signup process is unavailable.
Source :https://ns1.com/signup?plan=startup


ripienaar:
thank you

PR #3024, opened 2023-08-14
remove airbrake

moonsun1987:
looks like it requires credit card information to sign up now
image

## Requirements


ripienaar:
Requiring a CC isnt a problem - but indeed seems they have no more free tier.

PR #3041, opened 2023-08-29
remove humio.com

moonsun1987:
no free tier as far as I can tell, humio.com redirects to falcon logscale which only offers a free trial that you can "request" so likely a manual approval is needed even for a time boxed trial. 

301 moved permanently

> humio.com — Free up to 16 GB/day with 7 days retention

## Requirements


ripienaar:
Thanks

PR #3042, opened 2023-08-30
remove freegeoip.app and add ipbase.com

ItsNiceCraft:
## Requirements

[freegeoip.app](https://freegeoip.app) now redirects to [ipbase.com](https://ipbase.com).
According to the internet archive, ipbase most likely bought the domain after freegeoip went down.

This PR removes freegeoip and adds ipbase.


ripienaar:
Thank you

PR #3047, opened 2023-09-04
update connectycube ,remove push technology

mzltest:
1. https://connectycube.com/pricing/ its 1000 users now not 10k mau
image

2. push technology now redirects to https://www.diffusiondata.com/ and it no longer has a free plan
image


ripienaar:
Thanks, both of those seem ok - though I am reluctant to add dolby as they already killed one free tier

PR #3049, opened 2023-09-06
remove macrometa

mzltest:
only 30day free trial available now i think  , https://www.macrometa.com/pricing

image
image

according to the support there *was* a playground plan available but its no longer accepting signups
https://support.macrometa.com/hc/en-us/articles/10368164728973-Join-the-Playground
(it automatically redirects to the trial signup page)
image


ripienaar:
Thanks

PR #3053, opened 2023-09-14
propose dappling network

Bookcliff:
Propose [dAppling Network](https://dappling.network) - a free web hosting service that makes it quick and easy to host static sites on IPFS

[docs](https://docs.dappling.network)

Thanks for reviewing :))

## Requirements


ripienaar:
Thank you

PR #3058, opened 2023-09-17
Add ethereal.email service to the list

SDemonUA:
Ethereal is a fake SMTP service, mostly aimed at Nodemailer and EmailEngine users (but not limited to). It's a completely free anti-transactional email service where messages never get delivered. https://ethereal.email

## Requirements


ripienaar:
Thank you

PR #3060, opened 2023-09-17
Add Reshaped to the Design and UI section

blvdmitry:
## Requirements


README.md
```diff
@@ -1442,6 +1442,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+
   * [Glyphs](https://glyphs.fyi/) -- Free, The Mightiest Icons on the Web, Fully editable & truly open source design system.
   * [ShadcnUI](https://ui.shadcn.com/) -- Beautifully designed components that you can copy and paste into your apps. Accessible. Customizable. Open Source.
   * [HyperUI](https://www.hyperui.dev/) -- Free Open Source Tailwind CSS Components.
+  * [Reshaped](https://reshaped.so) -- Free React library. Professionally crafted design system in React & Figma for building products of any scale and complexity.
```
ripienaar, line 1445: This is a bit too focussed for this list I think - these component packs are in general a bit on the edge of what I think should be listed here, but this one more so as its quite specifically targetted as I understand

blvdmitry: In general I find it somewhat close to the flexibility of other UI libraries mentioned here but there is definitely a difference in between using things built on top of tailwind and not in terms of how much you can change on your side afterwards. So totally getting your point and feel free to close the PR if you think it's not the right fit

PR #3069, opened 2023-09-18
docs: adds grep.app under code search and browsing

cristianoliveira:
- A free tool for performing advanced query through half million git repos.

## Requirements


README.md
```diff
@@ -524,6 +524,7 @@ This list is the result of Pull Requests, reviews, ideas and work done by 1100+
   * [sourcegraph.com](https://about.sourcegraph.com/) — Java, Go, Python, Node.js, etc., code search/cross-references, free for Open Source
   * [tickgit.com](https://www.tickgit.com/) — Surfaces `TODO` comments (and other markers) to identify areas of code worth returning to for improvement.
   * [CodeKeep](https://codekeep.io) - Google Keep for Code Snippets. Organize,Discover and share code snippets, featuring a powerful code screenshot tool with preset templates and linking feature.
+  * [grep.app](https://grep.app) - Search across a half million git repos. It allows you to use advanced searching backed by elasticsearch.
```
ripienaar, line 527: Seems nice, but why not just use github search? Is it just because of the supported regex?

cristianoliveira: So grep.app is a better version of https://searchcode.com/ I tried to use that but it seems isn't up-to-date. It also has some advanced querying and previews which GitHub search doesn't have

ripienaar: Github does previews just fine, I am not against adding this - its just seems a bit like its not adding significant value over what exist already.  Main additions seem to be regex and case sensitivity?

Screenshot 2023-09-19 at 10 41 23

PR #3085, opened 2023-09-19
Add Embrace to Crash and Exception Handling section 

contrecc:
Happy to add Embrace to Free for Developers. Mobile developers can get full access to our Pro tier free for 45 days, and then we have a Free plan up to 1 million user sessions every year.

## Requirements


ripienaar:
Thank you

PR #3088, opened 2023-09-19
update service restrictions

mbfrias:
forwardemail.net also requires paid plans for a .lat domain
Screenshot 2023-09-19 at 13 56 21


ripienaar:
thanks

PR #3094, opened 2023-09-19
Fixes feature toggles section formatting

JakeSteam:
Fixes formatting of the "Feature Toggles Management Platforms" section.

As can be seen on the current live site, the bullet point formatting is slightly off (1 too many spaces before each item), resulting in:
* Vertical space between items
* Broken formatting on the "Back to top" link
* The "Fonts" section being entirely indented

image


ripienaar:
Thank you!

PR #3102, opened 2023-09-21
Add prompt terminal

tracymiranda:
## Requirements


README.md
```diff
@@ -1577,6 +1577,7 @@ Update Time, 5 active automations, webhooks.
    * [Mocklets](https://mocklets.com) - a HTTP-based mock API simulator, which helps simulate APIs for faster parallel development and more comprehensive testing, with lifetime free tier.
    * [Paiza](https://paiza.cloud/en/) — Develop Web apps in Browser without having the need to setup anything. Free Plan offers 1 server with 24 hours lifetime and 4 hours running time per day with 2 CPU cores, 2 GB RAM and 1 GB storage.
    * [Prepros](https://prepros.io/) - Prepros can compile Sass, Less, Stylus, Pug/Jade, Haml, Slim, CoffeeScript and TypeScript out of the box, reloads your browsers and makes it really easy to develop & test your websites so you can focus on making them perfect. You can also add your own tools with just a few clicks.
+   * [Prompt](https://commandline.dev/) - Prompt is the modern terminal for 10x devs. Render anything inline. Save sessions and history. Powered by open web standards.
```
ripienaar, line 1580: Not sure we have any other terminals on this list, kind of doesnt fit the bill - only client side stuff we tend to list are IDEs.  If there is already another terminal we can add it

WalysonGO: Only **MacOS**

tracymiranda: Updated to specify MacOS only (for now). 
Is there a different category to add it in? There are other similar types of modern code-editing terminals not in the list (e.g. warp.dev).

ripienaar: If you have 3 or 4 to add you can make a category else this is fine, let me know

tracymiranda: Not 3 or 4 just yet (will keep an eye on it - expect it will be like that in a year) - so for now please can we stick with this category, thx.

PR #3109, opened 2023-09-24
Add deco.cx — edge app platform for instant ecommerce frontends

guifromrio:
Amazing list, thanks for putting this together. Hope to see our free tier put to good use! 

Cheers from Brazil 🇧🇷 

## Why we built deco.cx

We were frustrated about how hard it is to deploy a modern TS + Tailwind website with a proper visual editor/admin UI for business users. **So many choices to make!** Which framework? Which platform to deploy on? Which headless CMS? Which A/B testing tool? Which monitoring service? How do we wire all of that together? Oh no, there goes the budget again...

That's why we built deco.cx: so that anyone with basic TypeScript and CSS can deploy a fully-functional, enterprise-grade storefront connected to any API as a data source within a few minutes, AND get PageSpeed 90+. 

We're an edge-native frontend platform focused on making ecommerce websites fast. You write TSX + Tailwind, we generate a beautiful visual CMS automatically. We render your app at the edge, which allows for fast *and* personalized sites at scale, so you can use our built-in A/B testing and content segmentation without fear. Built by devs with 10+ years experience scaling global enterprise ecommerce web properties. 

## Requirements


ripienaar:
thank you

PR #3110, opened 2023-09-25
docs: remove Scaleway. free 75GB storage not available anymore

peakxl:
The 75GB free object storage is no longer offered by Scaleway. Only the 75GB free egress each month remains. I suggest deleting Scaleway from IaaS altogether.

No longer on the website:
image

Confirmation from support:
image


Ardakilic:
For Googlers who came here:

My invoice is not yet populated with this offer removed. I should have seen some increase, but yet no action there. I'll post updates if it's being populated.

I believe they did this possibly because of Mastodon. Mastodon 4.2.0 introduced storage class option for storage, and people would use their system and store up to 75GB of assets as `ONEZONE_IA` for free.

PR #3118, opened 2023-09-29
Biggest amount of Grammar changes

3kh0:
I just sat down for 2 hours and did it all. I think that is all of it!


3kh0:
LOL, the diff is so big, it will not even render on the website

3kh0:
Just saw a mistake

PR #3138, opened 2023-10-13
Update README.md

RymMichaut:
Adding Taipy Cloud to the list

## Requirements


README.md
```diff
@@ -1174,6 +1174,7 @@ Update Time, five active automations, webhooks.
   * [SourceForge](https://sourceforge.net/) — Find, Create, and Publish Open Source software for free
   * [Stormkit](https://stormkit.io/) — Integrate building, deploying, and hosting seamlessly with the git flow of your JAMStack or Node.JS app. 50 GB bandwidth and 10m monthly requests for free, including free SSL.
   * [surge.sh](https://surge.sh/) — Static web publishing for Front-End developers. Unlimited sites with custom domain support
+  * [Taipy Cloud](https://www.taipy.io/get-started-with-taipy-cloud/) - Provides the most suitable cloud tool to host, deploy, and share your Taipy applications easily. In addition, this platform provides the ability to manage, store, and maintain the various states of your backend.
```
ripienaar, line 1177: generally we try to avoid this kind of thing where its one tech only - like its a cloud to host apps written in one specific python lib.  That seems too specific for a generic list like this.

PR #3147, opened 2023-10-22
Added the IP API named ipapi.is to the list

NikolaiT:
I added the TLD in the title, since it would have been a duplicate otherwise. I hope that is okay.

## Requirements


FatGrizzly:
Thanks

PR #3172, opened 2023-11-17
remove sauce labs

mzltest:
https://saucelabs.com/pricing
sauce labs no longer offers free plan(only 28 day free trial),which is also stated in the email they set to current free plan users:
image


ripienaar:
Thank you

PR #3179, opened 2023-11-24
Update Prompt to Wave Terminal

tracymiranda:
Updates to reflect that Prompt changed name to Wave Terminal and now works on Linux.


ripienaar:
Thanks!

PR #3181, opened 2023-11-27
Correct grammar error

Eason-S-Lu:
Update _one days_ to _one day_ in [Log Management](https://free-for.dev/#/?id=log-management) section


ripienaar:
Thank you

PR #3188, opened 2023-12-06
Remove elephantsql from Managed Data Services

sahwho:
elephantsql has discontinued their free tier plan ([source](https://www.elephantsql.com/blog/turtle-discontinued.html)). As such, it should be removed from this list.


ripienaar:
Thank you

sahwho:
Welp, it looks like they've since reverted that decision! The (old link)[https://www.elephantsql.com/blog/turtle-discontinued.html] returns a 404 and the (pricing page)[https://www.elephantsql.com/plans.html] now lists Tiny Turtle as an option (same config, 20 MB, 5 connections). This PR can probably just be reverted if there's no merge conflicts.

ripienaar:
As per the PR:

> Ok well, i dont really have time for the ping pongs - they'll probably kill it again future, so I am happy to leave it off for now.

PR #3189, opened 2023-12-08
Update AWS SES free tier limits

levnovikov:
AWS SES limit was changed from 62000 emails per month to 3000 emails: https://aws.amazon.com/ses/pricing/


ripienaar:
thank you

PR #3197, opened 2023-12-15
Removed MeiliSearch from Searches. No longer free

RayyanNafees:
Removed **MeiliSearch** entry from the [Search](https://free-for.dev/#/?id=search) section since it's no longer free


ripienaar:
Thanks

PR #3202, opened 2023-12-21
Remove 4 Broken List from the List- docsapp.io ,www.3dmodelscc0.com, www.selefra.io and cognalys.com

madhuryadutta:
**The URL of the following Services are broken/unavailable** -----

- docsapp.io
docs

- www.3dmodelscc0.com
3dmodel

- www.selefra.io 
selefra

- cognalys.com
cogna


ripienaar:
Thank you

PR #3204, opened 2023-12-23
Remove PageCDN from the list (CDN and Protection)

madhuryadutta:
**[Pagecdn.com ](https://pagecdn.com/)is returning a Cloudways 403 error Page**

Screenshot from 2023-12-23 13-57-48


ripienaar:
thank you

PR #3210, opened 2023-12-29
Remove Stormkit

1toldyou:
No longer offer free tier, became free trial instead

https://stormkit.io/blog/faq#:~:text=Why%20there%20is%20no%20free%20tier%3F


ripienaar:
Thank you

PR #2117, opened 2021-09-19
Add DynaPictures

alexzender:

PR #3213, opened 2024-01-05
Add Templated.io to APIs

camposped:
## Requirements


README.md
```diff
@@ -220,6 +220,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1100+ peo
   * [Apify](https://www.apify.com/) — Web scraping and automation platform to create an API for any website and extract data. Ready-made scrapers, integrated proxies, and custom solutions. Free plan with $5 platform credits included every month.
   * [API Mocha](https://apimocha.com) - Completely free online API mocking for testing and prototyping.  Make up to 500 daily requests, fully customizable API responses, and download mock rules as a Postman collection.
   * [APITemplate.io](https://apitemplate.io) - Auto-generate images and PDF documents with a simple API or automation tools like Zapier & Airtable. No CSS/HTML is required. The free plan comes with 50 images/month and three templates.
+  * [Templated](https://templated.io) - Automate your images with a simple API and no-code integrations. Design your template in a Canva-like aditor and render images (.jpg, .png, .webp) and PDFs with code. Free tier plan with 30 credits per month.
```
ripienaar, line 223: What does 1 credit get you? If its 1 render:1 credit then I dont think 30 is a useful number, I can easily see consuming that just during basic dev cycles

camposped: It's an estimate of 1 image/pdf per day. The credit resets every first day of the month. 
Usual use cases for that are e-commerces that need to auto generate a banner every day or a blog that needs a post cover on every new post.

The tool is completely accessible after the credits ends and I'm not blocking the user api key right away after he runs out of credit but only sending him a notice email with a grace period.

ripienaar: With such a low limit developers couldnt build against this without going out of pocket, couldnt run CI jobs etc, couldnt properly explore or integrate the service.

The purpose of the list - as per the top of the readme - is to promote services that make such activities easy.

So I am afraid I am going to reject this one thanks.

camposped: I’m going to increase the monthly quota to 100. Do you think that’s enough for initial assessment and integration with the API?

There are a few similar solutions in the list already that offer fewer credits than that.

ripienaar: Yeah sounds good. 

aa the list has grown huge I have become a bit more strict. You can PR to remove the ones with very low limits.

camposped: @ripienaar I've increased the quota on our system and the info is now updated in the PR.

ripienaar:
Thank you

alexzender:
Hi @ripienaar 
sorry to see our DynaPictures API is removed from the list. We were added in 2021 - https://github.com/ripienaar/free-for-dev/pull/2117 and serve developers devotedly since then.

Would appreciate if you could reconsider your decision.

Thanks,
Alex

ripienaar:
@alexzender that looks like unintentional, I cant see why it was removed.

I can see though that templates.io have apparently removed their free tier and now offer just a free trield, I should have known it was a scam when the initial offering was so useless. 

I'' revert this pr

ripienaar:
Actually on second thought I think the limit dynapictures offer is too low to be useful @alexzender. The purpose here is to allow developers to do their work without cost, you cant develop an integration with so few free images nor can you even build a CI system.  It'll stay off I am afraid.

alexzender:
Understood... We have a plenty of users on a free tier though, and there are other features like media library, where image generation credits are not used.

Thanks for reconsideration and the quick feedback though.

PR #3214, opened 2024-01-07
Add Data Dead Drop to APIs & Data

hschne:
Data Dead Drop provides anonymous file sharing for small files. Useful for CIs or just getting data from one machine to another if you have nothing but cURL and an internet connection. 

## Requirements

 * [x] It has a free tier not just a free trial. 


ripienaar:
thanks

PR #3234, opened 2024-01-31
Add MoniForm

Barctic:
## Requirements


ripienaar:
Yes this is pointless. Looks like a free tier designed to make it on the list rather than be of any actual value.

Barctic:
Sorry you feel that way. We have worked with many small businesses that went multiple months without knowing their contact form was not working (either their mail server was misconfigured, captcha not working, or form itself not working well, etc). They resulted in 'checking their own forms' AKA emailing themselves every 6 months to make sure it is indeed working. Also, nothing needs to be installed, and works on any contact form. Appreciate the feedback though, maybe our value prob could be strengthened.

ripienaar:
With your twice a year check frequency they would still go multiple months without knowing. 

This is a useful service to be clear. The free tier isn’t.

Barctic:
I see your point. Maybe we need to split any fully automated tests (ie: 404s) from anything that could have a manual component (filling out a captcha) to strengthen the offering. Obviously anything automated, its trivial to increase the frequency of checks for free users. Although it could reduce the total value of each test for free. Again, thanks for the comments.

PR #3240, opened 2024-02-05
Add NotificationAPI to Messaging

sahandseifi:
Thanks for reviewing!

PS. I noticed the Messaging and Streaming list does not follow any particular order. Let me know if you need any help with that.

## Requirements


ripienaar:
Great, generally it would be nice to see the sections sorted but I cant exactly expect people who wanted to spend 2 minutes adding an entry to keep sorting things, so inevitably it drifts into chaos every now and then :)

You're welcome to sort it if you're so inclined

ripienaar:
I've merged this one, feel free to do followups about sorting

PR #3241, opened 2024-02-06
Added ArticlePlanet

SH20RAJ:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1763,6 +1763,25 @@ Update Time, five active automations, webhooks.
   * [Hashnode](https://hashnode.com/) — Hassle-free Blogging Software for Developers!.
   * [Medium](https://medium.com/) — Get more thoughtful about what matters to you.
   * [AyeDot](https://ayedot.com/) — Share your ideas, knowledge, and stories with the world for Free in the form of Modern multimedia short-format Miniblogs.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1763,6 +1763,25 @@ Update Time, five active automations, webhooks.
   * [Hashnode](https://hashnode.com/) — Hassle-free Blogging Software for Developers!.
   * [Medium](https://medium.com/) — Get more thoughtful about what matters to you.
   * [AyeDot](https://ayedot.com/) — Share your ideas, knowledge, and stories with the world for Free in the form of Modern multimedia short-format Miniblogs.
+  * [ArticlePlanet](https://articleplanet.vercel.app/) - An open platform for users to publish articles on any topic, connect with other writers, and build an audience. It aims to democratize content creation. Save data on Gists
+  * - Write articles on any subject you want
```
ripienaar, line 1767: As you can see from other entries on the list, they do not look like this, they are short, sumaries and cover the important aspects of the service. Please create a entry that fit the style

ripienaar: OK, then perhaps its also just not fit for the list if it cant be described.

Also on closer investigation, there is only one user at the moment.

So I'll reject this for now until its a bit more popular - and so likely useful for the purpose of this list.

SH20RAJ: ma chuda

PR #3246, opened 2024-02-12
added: VIM in IDE & Code Editing section

gambhirsharma:
## Requirements


README.md
```diff
@@ -1632,6 +1632,7 @@ Update Time, five active automations, webhooks.
    * [SoloLearn](https://code.sololearn.com) — A cloud programming playground well-suited for running code snippets. Supports various programming languages. No registration is required for running code, but it is necessary when saving code on their platform. Also offers free courses for beginners and intermediate-level coders.
    * [stackblitz.com](https://stackblitz.com/) — Online/Cloud Code IDE to create, edit, & deploy full-stack apps. Support any popular NodeJs-based frontend & backend frameworks. Shortlink to create a new project: [https://node.new](https://node.new).
    * [Sublime Text](https://www.sublimetext.com/) - Sublime Text is a popular, versatile, and highly customizable text editor used for coding and text editing tasks.
+   * [VIM](https://www.vim.org/) - Vim is a highly configurable text editor built to make creating and changing any kind of text very efficient. It is included as "vi" with most UNIX systems and with Apple OS X.
```
ripienaar, line 1635: while yes its an IDE, we're not listing every piece of open source software here the list would be enormous if we did, vim is everywhere.  So we focus on thins that isnt already on every computer.

Thanks!

PR #3267, opened 2024-02-21
Added deta.space in Web Hosting Section

vulrun:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1213,6 +1213,7 @@ Update Time, five active automations, webhooks.
   * [MDB GO](https://mdbgo.com/) - Free hosting for one project with two weeks Container TTL, 500 MB RAM per project, SFTP - 1G disk space.
   * [Patr Cloud](https://patr.cloud/) — An easy-to-use cloud platform, among its paid services it offers to host three static sites for free.
   * [Serv00.com](https://serv00.com/) — 3 GB of free web hosting with daily backups (7 days). Support: Crontab jobs, SSH access, repositories (GIT, SVN, and Mercurial), support: MySQL, PostgreSQL, MongoDB, PHP, Node.js, Python, Ruby, Java, Perl, TCL/TK, Lua, Erlang, Rust, Pascal, C, C++, D, R, and many more.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1213,6 +1213,7 @@ Update Time, five active automations, webhooks.
   * [MDB GO](https://mdbgo.com/) - Free hosting for one project with two weeks Container TTL, 500 MB RAM per project, SFTP - 1G disk space.
   * [Patr Cloud](https://patr.cloud/) — An easy-to-use cloud platform, among its paid services it offers to host three static sites for free.
   * [Serv00.com](https://serv00.com/) — 3 GB of free web hosting with daily backups (7 days). Support: Crontab jobs, SSH access, repositories (GIT, SVN, and Mercurial), support: MySQL, PostgreSQL, MongoDB, PHP, Node.js, Python, Ruby, Java, Perl, TCL/TK, Lua, Erlang, Rust, Pascal, C, C++, D, R, and many more.
+  * [Deta Space](https://deta.space/) - Unlimited apps with scheduled actions, an unlimited number of custom domains with SSL support, 512MB RAM allocation, and read-only filesystems within sandboxed Linux VMs. Users can securely use unlimited storage and database capabilities in their self-created systems, akin to Firestore.
```
ripienaar, line 1216: early access, alphas, services with no future commitment to free tiers etc are not allowed on the list

vulrun: [Pricing](https://deta.space/pricing) link was previously displayed on the homepage and has been revised as of December 2023. Related to what they offer is documented [here](https://deta.space/docs/en/build/reference/runtime).

ripienaar: they dont even link to the docs from the home page and signup criteria is "apply" etc, sorry no just too much effort to confirm this is legit and keep the index current, so I'd rather just not take such low quality web sites

vulrun: @ripienaar  No problem at all, just discovered a great platform for app deployment and was excited to share it worldwide. They've recently updated the signup process to resemble surveys might be aiming to understand their audience better.

ripienaar: No problem, appreciate the effort to try and add it - but ultimately we need to consider the list health also and sometimes have to just say no :(

PR #3271, opened 2024-02-23
added aceternity.ui in Design and UI

mxnan:
added aceternity.ui in Design and UI

## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1445,7 +1445,7 @@ Update Time, five active automations, webhooks.
 **[⬆️ Back to Top](#table-of-contents)**
 
 ## Design and UI
-
+  * [Aceternity UI](https://ui.aceternity.com/) - Beautiful components built on framer motion , copy and paste .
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1445,7 +1445,7 @@ Update Time, five active automations, webhooks.
 **[⬆️ Back to Top](#table-of-contents)**
 
 ## Design and UI
-
+  * [Aceternity UI](https://ui.aceternity.com/) - Beautiful components built on framer motion , copy and paste .
```
ripienaar, line 1448: As you can see from other entries here they tend to list what is free etc.  Also please keep formatting such as empty lines consistent.

Regarding this site, I do not see any copyright notices for these components that are free, are they all original works by aceternity or what are their origin?  Is it just an index of things found or all original works?

mxnan: yes aceternity ui is free, but some features paid , all works by aceternity and their owner this repo will have use of this

ripienaar: Happy to add it when thats clearly stated

PR #3281, opened 2024-02-28
Remove Fly.io from Readme

altuzar:
## Requirements

Fly.io has no free tier anymore :( 

Starting since January 2024, all new accounts receive a $5 credit and require credit card to deploy. Then when the credit is used, the account is moved to the Hobby plan, that costs $5 + usage. 

Found about this in the hard way hehe. Great Heroku alternative otherwise.

https://fly.io/plans

From https://fly.io/docs/about/pricing/

Screen Shot 2024-02-28 at 4 08 25 PM


ripienaar:
Thank you

unixfox:
Technically it's still free "forever" unless you use all the $5. Which means you have to only create apps in the Free allowances limits: https://fly.io/docs/about/pricing/#free-allowances

I vote to add it back to free-for-dev, but with a big warning.

> When you sign up for a Fly.io account, you get a one-time $5 free trial credit to let you test-drive Fly.io at no cost. **You won’t be charged for the $5/month Hobby plan subscription until the credit has been used up**. The free trial credit doesn’t expire.

> Usage **beyond our [free resource allowances](https://fly.io/docs/about/pricing/#free-allowances)** consumes your free trial credit.

> The free trial credit applies to your default (“personal”) organization that we create for you on sign-up. **When the free trial credit is used up**, that organization is automatically placed on the $5/month [Hobby plan](https://fly.io/plans) (we’ll send you an email when it happens).

ripienaar:
They just make it too complex :) I have to consider the maintainability of the list long term and these kinds of overly restrictive or overly complex hoops to access the free resources just doesn’t add value.

imba-tjd:
new users can't register *Hobby* plan.
The *Pay As You Go* plan is $0 + **Usage**.
> There are no free allowances on the Pay As You Go plan

So the conclusion is no free plan.

PR #3312, opened 2024-03-22
Update README.md

flxg:
## Requirements

[ * ] This is Software as a Service not self hosted
[ * ] It has a free tier not just a free trial
[ * ] Pricing information is clearly visible without signup or phone calls
[ * ] The submission mentions what is free
[ * ] The submission is not already present in the list
[ * ] The service has contact details of those running it and a privacy policy

### DIFF
--- a/README.md
+++ b/README.md
@@ -636,7 +636,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1100+ peo
 
 ## Security and PKI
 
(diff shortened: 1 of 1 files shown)


FatGrizzly:
Maybe we can include its just for two users instead of a team, and maybe list a bit on whats free? (10 repos and 2 container images).

Thanks.

ripienaar:
Thanks

PR #3359, opened 2024-04-24
Remove Light CDN from the list

madhuryadutta:
The reason for this removal is that Light CDN has discontinued their 100 GB per month free tier plan. Instead, they now provide a 500 GB one-time package for 30 days.

**Sources:**
1. Notifications Page  
Link: https://www.lightcdn.com/en-US/notify/4e200ea42d2049a5b082a42ee41e7a75

image

2. New Pricing Page
Link: https://www.lightcdn.com/en-US/pricing

image


ripienaar:
Thank you

PR #3360, opened 2024-04-24
remove storj

rizkiv1:
No more free plan, changed into 1 month trial.
source: https://forum.storj.io/t/discontinuation-of-the-storj-free-tier/25332


ripienaar:
Thank you

PR #3361, opened 2024-04-25
Remove JumpCloud and Invision App

Themimitoof:
This PR removes the Jumpcloud link since they no longer offer a free tier. 

It also removes the Invision App Design link because they announced the end of life of Invision design  and will be effective at the end of the year:  https://www.invisionapp.com/inside-design/invision-design-collaboration-services-shutdown.


ripienaar:
Thank you

PR #3367, opened 2024-04-30
Adding Page Flows to "Design Inspiration" Section

kvenn:
I think I added all of the items in this section so I organized it a bit
* Update the "tags" to be consistent casing
* Alphabetical order

I also added a new one (PageFlows) which is pretty great and is almost entirely free.

## Requirements


ripienaar:
Thank you

PR #3368, opened 2024-04-30
add backgroundstlyler

Vijayabhaskar96:
Almost all features are free, except for a few customizable settings.

## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1824,6 +1824,7 @@ Update Time, five active automations, webhooks.
 
 ## Miscellaneous
 
+  * [BackgroundStyler.com](https://backgroundstyler.com) - Create aesthetic screenshots of your code, text or images to share on social media.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1824,6 +1824,7 @@ Update Time, five active automations, webhooks.
 
 ## Miscellaneous
 
+  * [BackgroundStyler.com](https://backgroundstyler.com) - Create aesthetic screenshots of your code, text or images to share on social media.
```
ripienaar, line 1827: Doesn't seem that useful, why would you use this and get a watermark when there are better free ones without a watermark.

Technically this is free but I am not just adding stuff, there needs to be a point.

Vijayabhaskar96: I'm the maker of the site. I simply thought it was a better alternative (it offers more styles, fonts, themes, frames, and syntax highlighting for over 100 languages) to many of the websites currently included in your repo, including one that also adds watermarks to the generated images. So I felt mine might fit in as well. If you find it spammy or redundant, close this PR, sorry.

ripienaar: Fair enough, I'll add it, but its not really differentiating from others in my mind

PR #3397, opened 2024-05-29
Add us.kg subdomain

neneeen:
## Requirements


README.md
```diff
@@ -1254,6 +1254,7 @@ Update Time, five active automations, webhooks.
 
   * [eu.org](https://nic.eu.org) — Free eu.org domain. The request is usually approved in 14 days.
   * [pp.ua](https://nic.ua/) — Free pp.ua subdomains.
+  * [us.kg](https://nic.us.kg/) - Free us.kg subdomains.
```
ripienaar, line 1257: Thanks, can give us some example domains you've registered we need to validate its not just a frontend to cloudflare etc.  To quality you MUST offer NS records to any nameserver.

PR #3398, opened 2024-05-29
Update Turso allowance

x-t:
For more, see

https://turso.tech/blog/turso-radically-increases-the-amount-of-databases-available-on-their-starter-and-scaler-plans-10a69ad94055

https://turso.tech/pricing


ripienaar:
Thank you

PR #3411, opened 2024-06-19
Added Hydrozen.io

raghuchinnannan:
Added Hydrozen.io to the list of monitoring tools. Hydrozen.io has a free forever plan which supports 10 uptime monitors, 5 heartbeat monitors, 1 Domain & SSL monitor, 1 DNS monitor and 1 Statuspage.

## Requirements


ripienaar:
Thanks

PR #3419, opened 2024-06-27
Add Gradient to the list

noashavit:
Adding Gradient to the list under the API, Data, and ML category

## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 ## APIs, Data, and ML
 
+  * [Gradient](https://synccomputing.com/) —  ML-Powered cluster management for Databricks. Gain full visibility into cost and performance of your databricks clusters, and reduce compute costs by 50% while meeting SLAs.
   * [Zerosheets](https://zerosheets.com) - Turn your Google Sheets spreadsheets into powerful API`s to rapidly develop prototypes, websites, apps and more. 500 requests per month available for free.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -209,6 +209,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 ## APIs, Data, and ML
 
+  * [Gradient](https://synccomputing.com/) —  ML-Powered cluster management for Databricks. Gain full visibility into cost and performance of your databricks clusters, and reduce compute costs by 50% while meeting SLAs. The free plan includes $100 free core hours.
```
ripienaar, line 212: is this $100/monht or one off?

noashavit: yeah it's confusing. I'll edit it to be clearer - it's 100$ worth of core hours, but I think I can make that clearer.

ripienaar: Ok. It same question. Is this per month or one off?

in other words is it n core hours a month for free or use once then its done?

ripienaar: Ah I see that’s like 400+ days

ripienaar: Will have to go see what that means in real usable terms say I have a 2 node 8 core per server cluster

ripienaar: All of the site verbiage etc states this is a free trial not a free tier and indeed the capped hours nature of this seem to concur.

Quoting the top of the readme:

```
 To be eligible, a service must offer a free tier, not just a free trial. The free tier must be for at least a year if it is time-bucketed. 
```

So I'll go ahead and cose this as not suited

PR #3423, opened 2024-07-06
added openBlog

samuelail:
## Requirements


README.md
```diff
@@ -1802,6 +1802,7 @@ Update Time, five active automations, webhooks.
   * [Hashnode](https://hashnode.com/) — Hassle-free Blogging Software for Developers!.
   * [Medium](https://medium.com/) — Get more thoughtful about what matters to you.
   * [AyeDot](https://ayedot.com/) — Share your ideas, knowledge, and stories with the world for Free in the form of Modern multimedia short-format Miniblogs.
+  * [OpenBlog](https://openblog.dev) - Free and open blogging app for developers, powered by Markdown.  You don't even need to have an account to use it.
```
ripienaar, line 1805: How’s this better than just sticking md on gist?

it’s worse in many ways since there is no attribution or ability to edit etc. 

As it is, I just don’t think it’s worth adding to the list maybe it’s a bit early days.

PR #3434, opened 2024-07-20
Two more DDNS services

ChrSteinert:
I added two no dynamic DNS services.
Especially the fine grained management capabilities of Dynv6 are pretty rare, and (as far as I can see) currently with no service on the list.

Noip.at on the other hand is super simple – not even having a login. Register a domain and receive a token to update it. It is missing more management features though.

## Requirements


FatGrizzly:
thanks

PR #3447, opened 2024-07-30
Update README.md

romanfurman6:


README.md
```diff
@@ -1478,6 +1478,7 @@ Update Time, five active automations, webhooks.
   * [degoo.com](https://degoo.com/) – AI based cloud storage with free up to 20 GB, three devices, 5 GB referral bonus (90 days account inactivity).
   * [MConverter.eu](https://mconverter.eu/) – Convert files in bulk. Supports many file formats, including new ones like [AVIF](https://mconverter.eu/convert/to/avif/). Extract all image frames from videos. Free for up to ten 100MB-files per day, processed in batches of two.
   * [ImageEngine](https://imageengine.io/) – ImageEngine is an easy to use global image CDN. Sub 60 sec setup. AVIF and JPEGXL support, WordPress-, Magento-, React-, Vue- plugins and more. Claim your free developer account [here](https://imageengine.io/developer-program/).
+  * [DocsParse](https://docsparse.com/) – GPT powered AI processing of PDFs, Images, into structured data in JSON, CSV, EXCEL formats. 30 credits for free each month.
```
ripienaar, line 1481: The site does not say 30/month 

Screenshot 2024-07-31 at 07 24 19

Just that 30 credits are given but doesn't mention a frequency, does say its a "trial" though.

It also doesn't define what a "credit" is or what can be done with it

romanfurman6: Hey, ye sorry we've fixed FAQ. It's not technically Trial but Tier, Trial more in terms of testing product. You can check it out inside of the app, mb it will be more clear

ripienaar: Thanks looks good. 1550 entries on the list, I am not signing up to each to see prices :)

PR #3448, opened 2024-07-31
Add freeimage.host to "Storage and Media processing"

OgbeniHMMD:
## Requirements


gshapiro:
Note your commit says, "128 MB limit" while the web site says, "64 MB limit".

ripienaar:
Oh sorry, I only saw the comment after - please send a PR to fix that else I'll revert tomorrow

PR #3453, opened 2024-08-07
Update README.md

imabutahersiddik:
Addition:
Category: Website Hosting
Name: Erath
Description: Erath offers free web page hosting through an innovative storage-less webpage hosting model. By embedding data directly into URLs, Erath ensures secure and immutable access to web content. With AES-256 encryption, it guarantees the protection and integrity of your pages while providing perpetual availability without the need for traditional data storage solutions. Perfect for developers looking to explore cutting-edge web hosting technologies.

## Requirements


ripienaar:
while this is interesting I am not convinced it really is something for the list, perhaps if it wasn't just some vercell app but more mature but even then - I just dont think it works for the goal of the list sorry

PR #3461, opened 2024-08-19
Update README.md

hansanaD:
Added **getfreedomain.name** to the list.

## Requirements


README.md
```diff
@@ -1278,6 +1278,7 @@ Update Time, five active automations, webhooks.
   * [eu.org](https://nic.eu.org) — Free eu.org domain. The request is usually approved in 14 days.
   * [pp.ua](https://nic.ua/) — Free pp.ua subdomains.
   * [us.kg](https://nic.us.kg/) - Free us.kg subdomains.
+  * [getfreedomain.name](https://www.getfreedomain.name/) - Free domain names and free short subdomains with no ads & full DNS support!
```
ripienaar, line 1281: It doesn’t seem to me that this is a service pricing domain names rather it’s an index to other services so I think this description is misleading. 

Since we don’t add dns services here as a category I also don’t think I will list a index of tbem sorry

PR #3469, opened 2024-08-28
Added a new section for LLM Observability

LinaLam:
- Added Helicone in the section

## Requirements


README.md
```diff
@@ -770,6 +771,12 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 **[⬆️ Back to Top](#table-of-contents)**
 
+## LLM Observability
+
+  * [Helicone.ai](https://helicone.ai/) — Open-source LangSmith alternaive for logging, debugging and improving your AI apps. Free up to 100k logs/month.
```
ripienaar, line 776: Thank you. We do not add 1 entry sections so if there are others that fit the bill you can move them here. Else you have to add it in an existing section.

LinaLam: thanks so much for getting back - just made a new PR with a few more companies in this category :) 

will close this one

PR #3476, opened 2024-09-03
Added 2 free temp email services that are entirely free

Tr0llie:
Added FakeMailGen https://www.fakemailgenerator.com/ to the list  and
added https://temp-mail.org/en/ to the list
Both Of these two services are entirely free and support many domains
both of these are not open source, use at your own risk. 

fakemailgen collects cookies which allows you to reload the site to view your inbox. It also collects some third party cookies for marketting purposes. It is not a privacy concern but, we can not know what they do with those data.

## Requirements

 * [ x] This is Software as a Service not self hosted
 * [ x] It has a free tier not just a free trial
 * [ x] Pricing information is clearly visible without signup or phone calls
 * [ x] The submission mentions what is free
 * [ x] The submission is not already present in the list
 * [ x] The service has contact details of those running it and a privacy policy


Tr0llie:
Found these services like 3 year ago and have been using them since, checked on this repo once in a while and didn't see  these in the list. would be great if it were added since i use them to signup for blogs or some articles to be able to read them.

ripienaar:
Thanks, these will be the last fake mail generators I accept, these things are all just derivative and don't add value one over the other, so I'll probably trim the list down also.

Tr0llie:
Yeah you might be right, those things are really useless sometime and does the same job.

PR #3478, opened 2024-09-04
Add Dynapik

ralphilius:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1500,6 +1500,7 @@ Update Time, five active automations, webhooks.
   * [coolors](https://coolors.co/) - Color palette generator. Free.
   * [Branition](https://branition.com/colors) - Hand-curated color pallets best fitted for brands.
   * [css-gradient.com](https://www.css-gradient.com/) - Free tool to quickly generate custom cross-browser CSS gradients. In RGB and HEX format.
+  * [Dynapik](https://dynapik.com) - Free media suite to quickly convert formats, remove backgrounds, resize and others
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1500,6 +1500,7 @@ Update Time, five active automations, webhooks.
   * [coolors](https://coolors.co/) - Color palette generator. Free.
   * [Branition](https://branition.com/colors) - Hand-curated color pallets best fitted for brands.
   * [css-gradient.com](https://www.css-gradient.com/) - Free tool to quickly generate custom cross-browser CSS gradients. In RGB and HEX format.
+  * [Dynapik](https://dynapik.com) - Free media suite to quickly convert formats, remove backgrounds, resize and others
```
ripienaar, line 1503: tried to remove some background, after minutes of sitting saying "removing background.." on one of the sample images nothing happens so I doubt this is ready for use

ripienaar: yeah I left it for like 20 minutes while I went to the shop :) think I'll pass for now on this one.

ralphilius: > tried to remove some background, after minutes of sitting saying "removing background.." on one of the sample images nothing happens so I doubt this is ready for use

This service run on your browser resource so maybe it's slow.

Tr0llie: > This service run on your browser resource so maybe it's slow.

i will test it out in firefox, since i was using brave browser for the first test.

ripienaar: I understand how it works, I was using a M2 Mac with safari, fairly standard setup.

Lets just skip it, there's no shortage of converters, we don't need add ones that are tempremantal

PR #3480, opened 2024-09-04
Added Typesense.org in "SEARCH" sector

Tr0llie:
https://typesense.org/
github - https://github.com/typesense/typesense
There are tons of live demo in the github repo, you should check them out. 

Only downside is that you need to selfhost typesense or use a free vps. But it doesn't use much and isn't resource intensive so it should works just fine. i am not sure if is "as fast as algolia" but it does seems "as fast" and its free so i give it a point just for being fast and for being free.

## Requirements


Tr0llie:
as stated "not for self-hosted software". I don't condone this PR to be merged since it kinda needs to be "self hosted" in a way

ripienaar:
You seem to realise it's not a fit, so why open the PR? :)

There are 1600+ watchers on this repo, all getting mails for PRs so it's worth keeping it short and on topic

Tr0llie:
> You seem to realise it's not a fit, so why open the PR? :)
> 
> There are 1600+ watchers on this repo, all getting mails for PRs so it's worth keeping it short and on topic

a chance is a chance. i will try to keep it short next time

PR #3492, opened 2024-09-11
Delete link invalid

leknoppix:
I have deleted a dead link.

the service https://onevm.uk.to/ is down


PlurrYT-Kolama:
Link invalid, they swapped to wiki page https://wiki.onevm.uk.to/en/start

PR #3494, opened 2024-09-12
add trtc(tencent rtc)

mzltest:
>[!IMPORTANT] 
> first year only offer(I already stated in the description) 
> In the bottom of the homepage https://trtc.io/ (that's very tricky since their signup/login page says free for unlimited time):
> image
> image

## Requirements

 https://trtc.io/pricing/rtc-engine


ripienaar:
Thank you

PR #3521, opened 2024-09-25
Remove services (Divio, EdgeOne) without free tier

axww:
According to NOTE "a service must offer a free tier, not just a free trial"
Divio provides 30 days trial
EdgeOne provides 14 days trail
Both of them have no free tier, so I removed them from the list.


ripienaar:
Thank you

PR #3524, opened 2024-09-27
Remove Metered.ca free offering

Mynacol:
Metered.ca reduced their free tier from 50GB TURN traffic per month to 500MB. See https://www.metered.ca/pricing

## Requirements


ripienaar:
This is a huge reduction and I wonder if it’s still even usable at that limit?

ripienaar:
Such a big drop in the offering I am inclined to just remove them. Would you mind updating your PR?

ripienaar:
Thank you

PR #3535, opened 2024-10-06
Added Heybounce

dorukcan:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -944,6 +944,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
   * [Emailvalidation.io](https://emailvalidation.io) - 100 free email verifications per month
   * [FakeMailGenerator.com](https://www.fakemailgenerator.com/) - A German Temporary / Disposable Mail generator. Support 10 domain, while giving you the freedom of creating unlimited addresses. (include ads) but other than that, there is no pricing included in the service, it is entirely free.
   * [forwardemail.net](https://forwardemail.net) — Free email forwarding for custom domains. Create and forward an unlimited amount of email addresses with your domain name (**note**: You must pay if you use .casa, .cf, .click, .email, .fit, .ga, .gdn, .gq, .lat, .loan, .london, .men, .ml, .pl, .rest, .ru, .tk, .top, .work TLDs due to spam)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -944,6 +944,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
   * [Emailvalidation.io](https://emailvalidation.io) - 100 free email verifications per month
   * [FakeMailGenerator.com](https://www.fakemailgenerator.com/) - A German Temporary / Disposable Mail generator. Support 10 domain, while giving you the freedom of creating unlimited addresses. (include ads) but other than that, there is no pricing included in the service, it is entirely free.
   * [forwardemail.net](https://forwardemail.net) — Free email forwarding for custom domains. Create and forward an unlimited amount of email addresses with your domain name (**note**: You must pay if you use .casa, .cf, .click, .email, .fit, .ga, .gdn, .gq, .lat, .loan, .london, .men, .ml, .pl, .rest, .ru, .tk, .top, .work TLDs due to spam)
+  * [Heybounce](https://www.heybounce.io/) - Fast, Accurate & Secure Email Validation Service. 100 free credits/month.
```
ripienaar, line 947: this appears to be a 100 credit one off trial, not a free tier. Even if it was a 100 credit a month free tier I wouldn't think that's an amount that would be useful to anyone. Wouldn't even be enough for a CI system.

dorukcan: Thanks for your time to review the pull request. 
I saw other email verification tools on the list that offer 100 credits per month and thought it would be a fit for the list. 
Also, i don't think that email verification services should be utilized in CI systems since there would be no transactional value in such integration (no email address?). This would be beneficial if you are implementing an email collection flow (sign-up, lead generation pages etc.) where you want to eliminate disposable, honeypot, or non-existent email addresses.

ripienaar: This service does not spell it out that it's 100/month.

dorukcan: Yeah you are right, my bad. Let me close the pr.

ripienaar: If the 100 was clarified as being each month and recurring, then sure, but that's not what it says. Can't exactly compare a service giving 100 credits with one giving 100 perpetually.

PR #3537, opened 2024-10-07
docs: add `una-ui` to the Design/UI section

phojie:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1581,6 +1581,8 @@ Update Time, five active automations, webhooks.
   * [Image BG Blurer](https://imagebgblurer.com/) -- Generate a blurred background frame for an image, using that image source as the background blur, for Notion, Trello, Jira, and more tools
   * [Webstudio](https://webstudio.is/) -- Open-source alternative to Webflow. The free plan offers unlimited websites on their domain. Five websites with custom domains. Ten thousand page views/month. 2 GB asset storage.
   * [Nappy](https://nappy.co/) -- Beautiful photos of Black and Brown people, for free. For commercial and personal use.
+  * [UnaUI](https://unaui.com/) -- The Atomic UI Framework for Nuxt, A technology that helps you build fast, modern, and lightweight web applications with the power of Unocss Engine.
+    
 
 **[⬆️ Back to Top](#table-of-contents)**
 


README.md
```diff
@@ -1581,6 +1581,8 @@ Update Time, five active automations, webhooks.
   * [Image BG Blurer](https://imagebgblurer.com/) -- Generate a blurred background frame for an image, using that image source as the background blur, for Notion, Trello, Jira, and more tools
   * [Webstudio](https://webstudio.is/) -- Open-source alternative to Webflow. The free plan offers unlimited websites on their domain. Five websites with custom domains. Ten thousand page views/month. 2 GB asset storage.
   * [Nappy](https://nappy.co/) -- Beautiful photos of Black and Brown people, for free. For commercial and personal use.
+  * [UnaUI](https://unaui.com/) -- The Atomic UI Framework for Nuxt, A technology that helps you build fast, modern, and lightweight web applications with the power of Unocss Engine.
```
ripienaar, line 1584: Thanks. I think this is too specific to a single technology - or worst a subset of a specific stack - to be applicable to a list of generic services and resources

phojie: No worries, So this is different case with the NextUI ?

phojie: UnaUI is an alternative to NextUI, btw ✅

ripienaar: Isn't it that UnaUI is specifically for Nuxt and that in turn is only for Vue? If that's the case then UnaUI is very specific, not just Vue specific but also onto of that Nuxt specific. Had it been for any Vue system I'd accept it.

Contract that with NextUI that's for any React app.

Well I am not a webdev but this is how I interpret the wording on the sites.

phojie: @ripienaar it support multiple framework that https://unocss.dev/integrations/ covered 🙂

ripienaar: I can't realistically deep dive into every one of the over thousand entries here to figure out their specific features and neither can the many others who maintain this list, based on this it is for Nuxt that's what I am going on.

Going by what's on unocss I am not even sure how you'd expect me to find the relation to some other unocss.dev site since it's not linked there nor are the docs linked to that site - I realise unaui used unocss as an engine but only because I spent way too much time digging into this already. 

Screenshot 2024-10-08 at 09 05 59

phojie: I understand 🙏, its a actually in WIP,. the documentations not even done.

PR #3546, opened 2024-10-20
Add WebTorrent trackers as general-purpose WebRTC signaling server

lessthanthree21:
## Requirements

Since free STUN servers are in the list, I believe this submission also satisfies the requirements.

## Rationale

For example, it could theoretically allow one to develop a multiplayer game without hosting a single server themselves.

---

The item can be worded a bit better, edit it as you like.


ripienaar:
Sorry, no I dont think this fits the purpose of the list

PR #3582, opened 2024-12-02
Add WeatherXu.com

StevenTso:
## Requirements

The [pricing page](https://weatherxu.com/pricing) has a free tier which is free for up to 10,000 API calls/month


ripienaar:
nice, thanks

PR #3585, opened 2024-12-03
Add MyStoryElf.com

afwcole:
## Requirements


afwcole:
May I ask why? Given other generative AI API products are a part of the list.

ripienaar:
Did you read the scope of the list and what it's about? Is it about kids?

afwcole:
> Did you read the scope of the list and what it's about? Is it about kids?

Ok fair enough

PR #3599, opened 2024-12-19
add Giscus

FireIsGood:
## Requirements

The contact details are on the organization (support@giscus.app) and the privacy policy is at [PRIVACY-POLICY.md](https://github.com/giscus/giscus/blob/main/PRIVACY-POLICY.md)

A service similar to Utternaces, though this uses Discussions which some may find more appealing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1839,6 +1839,7 @@ Update Time, five active automations, webhooks.
 ## Commenting Platforms
   * [GraphComment](https://graphcomment.com/) - GraphComment is a comments platform that helps you build an active community from the website’s audience.
   * [Utterances](https://utteranc.es/) - A lightweight comments widget built on GitHub issues. Use GitHub issues for blog comments, wiki pages, and more!
+  * [Giscus](https://giscus.app) - A comments system powered by GitHub Discussions. Heavily inspired by Utterances.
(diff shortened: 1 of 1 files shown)


Shahid439-max (approved):

README.md
```diff
@@ -1839,6 +1839,7 @@ Update Time, five active automations, webhooks.
 ## Commenting Platforms
   * [GraphComment](https://graphcomment.com/) - GraphComment is a comments platform that helps you build an active community from the website’s audience.
   * [Utterances](https://utteranc.es/) - A lightweight comments widget built on GitHub issues. Use GitHub issues for blog comments, wiki pages, and more!
+  * [Giscus](https://giscus.app) - A comments system powered by GitHub Discussions. Heavily inspired by Utterances.
```
ripienaar, line 1842: What's the SaaS here? You run it in your own GitHub Actions and GitHub is the backend - that means GitHub is the SaaS offering.

FireIsGood: Just looked into the specifics and it seems you're right. The app posts under the username "Giscis[bot]", similar to how Utterances has an [utterances bot](https://github.com/utterances-bot) though it is not an actual user.

Would the app still be similar enough to be included alongside Utterances? Or is this not technically applicable because of that difference?

ripienaar: Of another service here works the same it should probsbly be removed

PR #3603, opened 2024-12-28
add `is-a.dev`

wdhdev:
## Requirements

---

I'm aware you no longer allow free subdomain services, however I was hoping you would make an exception for is-a.dev as we are a well established subdomain service which has been around since 2020. We would also not fall under the verbatim copy paste rule either as we have our own custom made CI, with custom tests, all of which were made by myself.

We have over 3,000 users with over 5,500 subdomains. We are listed on Mozilla's [Public Suffix List](https://github.com/publicsuffix/list). We also have an open source software sponsorship from [Cloudflare](https://cloudflare.com) which provides us with their enterprise services.

Also, I think this is a good fit for the list as it is a short domain for developers, and is extremely popular. It is also worth noting that eu.org has not been active for around a year now, which has lead a few of their users to us.

I hope you can make an exception for us, if you cannot, I do understand. However I think listing us will benefit the majority of users visiting Free For Dev.

https://is-a.dev
https://github.com/is-a-dev/register

### DIFF
--- a/README.md
+++ b/README.md
@@ -1298,6 +1298,7 @@ Update Time, five active automations, webhooks.
 ## Domain
 
   * [eu.org](https://nic.eu.org) — Free eu.org domain. The request is usually approved in 14 days.
+  * [is-a.dev](https://is-a.dev) - Grab your own sweet-looking '.is-a.dev' subdomain.
   * [pp.ua](https://nic.ua/) — Free pp.ua subdomains.
   * [us.kg](https://nic.us.kg/) - Free us.kg subdomains.
 


ripienaar:
Sorry no, this was listed here before and removed.

wdhdev:
If I recall correctly, it was removed when you added the rule, however I think we should be eligible especially if us.kg (#3397) was added, which receives significantly more abuse and stuff than us. We also offer NS records, similar to eu.org and other listed services.

ripienaar:
That one wa kept because it has a good looking control panel offering an easy to use service. Owned by an established non profit organisation which at least ok the surface looks legit and have several impactful projects. In essence they took the time to build a professional NIC using tooling and UI that matches that of the industry. 

I don’t see how the level of abuse complaints a service like this gets is relevant - what would be relevant is evidence that they are not dealing with legitimate abuse complaints. I myself have worked on some of the first free domains ever and of course these services attract abuse - that’s a given it’s how it’s handled that matter. I don’t know how they handle their abuse but if you are going to come here and sling mud at other services please provide some supporting evidence of how they do not handle abuse. 

The reason these free domain services are banned is because of the amount of noise and effort they generate. Those running them are always here complaining, moaning about competitors, almost universally low quality GitHub based front ends to cloud flare, do not provide actual sub domains as they claim (I see you discouraging NS records) etc

wdhdev:
> I don’t see how the level of abuse complaints a service like this gets is relevant - what would be relevant is evidence that they are not dealing with legitimate abuse complaints. I myself have worked on some of the first free domains ever and of course these services attract abuse - that’s a given it’s how it’s handled that matter. I don’t know how they handle their abuse but if you are going to come here and sling mud at other services please provide some supporting evidence of how they do not handle abuse.

I noted the abuse with us.kg as the domain [was recently disabled](https://github.com/DigitalPlatDev/FreeDomain/blob/main/documents/announcement.md#december-25-2024) by the .kg registry for significant subdomain abuse (as far as I know) just a few days ago, I'm not trying to throw shame at them, just wanted to point it out that they might not be able to handle the abuse too much, compared to some other services. I do note their effort to help thwart it, however having an online dashboard, does allow much more abuse to flow through, than similar services on GitHub like js.org or us.

> That one wa kept because it has a good looking control panel offering an easy to use service. Owned by an established non profit organisation which at least ok the surface looks legit and have several impactful projects. In essence they took the time to build a professional NIC using tooling and UI that matches that of the industry.

For our website, we don't need it to be super fancy as we operate on GitHub mostly, and do not have a dashboard any more due to the abuse we received through it. Also notable, we have super minimal abuse, and any abuse we do find, is normally found extremely quickly and is removed. (We have removed multiple domains in the last few days alone.)

We mainly operate on GitHub because it is a well known service most developers already have accounts on, as we don't want people to have to create another account for something. GitHub Actions allows us to easily run [tests and validations](https://github.com/is-a-dev/register/tree/main/tests) in virtual environments, one of the main reasons why we use GitHub. Also it allows us to easily review new requests, without too much hassle. "Don't fix what ain't broke!"

> The reason these free domain services are banned is because of the amount of noise and effort they generate. Those running them are always here complaining, moaning about competitors, almost universally low quality GitHub based front ends to cloud flare, do not provide actual sub domains as they claim (I see you discouraging NS records) etc

We don't complain about any competitors, in fact we encourage them, we don't want the community to be limited to just us, we want them to have a choice. In fact, a lot of our maintainer team help maintain other smaller, free subdomain services.

We discourage NS records, but do not restrict them from users that need them. The reason we discourage them is that we already offer all the standard record types such as A, AAAA, CNAME, MX, TXT, etc, however we do, and will continue to allow NS records for users that do require them, if our supported record types are not enough, or if they need some features we don't offer (such as hiding their home IP behind a Cloudflare tunnel). We also support enterprise-level DDoS protection, through Cloudflare's proxy feature. We currently have 40 domains using NS records: https://github.com/search?q=repo%3Ais-a-dev%2Fregister+%22NS%22%3A+%5B&type=code Most of which were not paid and were approved by us as they were needed.

We never originally supported NS records for around 3 1/2 years, however that rarely didn't bother users and we still grew to over 4,500 domains before they were supported after our migration to Cloudflare.

We also have super active support on both GitHub and Discord, and nearly all valid requests are approved with-in 24 hours.

The reason I would like this service listed, is not for my own benefit, but for the benefit of the developer community who are looking for free subdomains.

ripienaar:
Again unsubstantiated claims about us.kg not handling abuse. If you have evidence of this please share. If it’s a real case of significant systemic abuse not being handled it seems unlikely to be fully resolved in one day - especially those days. 

This is exactly the kind of behaviour that got free domains banned and you are just continuing the cycle of free domain operators acting like children by throwing mud at each other but failing to offer any evidence to support the claims. 

The likelyhood of me ever accepting a service like yours have gone down rather quite significantly this morning. Suggest you stop replying now unless it is with substantiating evidence of these claims.

wdhdev:
Here are some illegal/disallowed websites using the `.us.kg` domain, which I found through a simple Google search, to support my earlier claims:
```
hXXps://corsflix.us.kg
hXXps://vipshop.us.kg
```
(both of which I will be reporting to the us.kg team)

The domain is also listed on Spamhaus' domain block list likely due to email abuse: https://check.spamhaus.org/results/?query=us.kg

The us.kg service does appear to be a really good service regardless, and I understand they may receive more abuse than some other services as they use a web dashboard, and I don't believe they manually review the domains. I would believe that they do try their hardest to prevent abuse though, as much as we do!

I'm not trying to argue my point across or throw shade at any services, sorry if I came across that way. I just personally think adding the domain would be beneficial to people viewing the list as it is good to have a variety of options, like you have with the other sections on the list, so I would hope you would reconsider 😄

ripienaar:
The question isn’t if free services are being abused. The question is are abuse being handled. We all know free services are subject to abuse which is why so many stop being free. 

So providing evidence of abuse is like evidence that the sky is blue. And entirely irrelevant to my points thus far. 

I can only assume you are being deliberate and refusing to answer the question which is proof of not handling abuse complaints. 

I will lock this now and block you.

PR #3607, opened 2025-01-03
Add Railway to the list of free tools and services

sheraz14:
This PR adds Railway to the list of free tools and services. It’s a modern platform for deploying apps and databases with a free tier. It was added to PaaS list.


ripienaar:
This seems like a free trial with $5 credit only?

PR #3619, opened 2025-02-09
Remove YugabyteDB

awinterstein:
There is no information about a free tier on their pricing page. See https://www.yugabyte.com/pricing/


ripienaar:
thank you

PR #3620, opened 2025-02-09
Update of conditions for Filess

awinterstein:
There are now two databases free, but with only 10 MB per database. See https://filess.io/#pricing


ripienaar:
damn, I tried to resolve the merge conflict and then messed up, would you mind rebasing and putting your change back in? Sorry about that! another PR edited the line above yours and sent GitHub in a conflict

PR #3628, opened 2025-02-17
Remove tab; Fix LocalOps placement;

localops-root:
Minor fix: Had an extra tab somehow in last commit. Removed it now.


ripienaar:
Thank you

PR #3629, opened 2025-02-17
Update README.md

ethansunray:
Add Temp Mail

### DIFF
--- a/README.md
+++ b/README.md
@@ -950,6 +950,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 ## Email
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -950,6 +950,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 ## Email
 
+  * [Temp Mail](https://tempmail.la) - Free Temporary Email, Protect Privacy, Avoid Spam Harassment
```
ripienaar, line 953: As per the notice shown when you opened the PR, we do not accept more of these.

```
```

PR #3641, opened 2025-02-25
Add core-web-vitals history checker

geeky-biz:
## Requirements

 * [x ] This is Software as a Service not self hosted

Note: This is completely free offering (nothing to pay for any amount of usage).

### DIFF
--- a/README.md
+++ b/README.md
@@ -848,6 +848,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
   * [appspector.com](https://appspector.com/) - Mission control for remote iOS/Android/Flutter debugging. Free for small traffic usage (64MB of logs).
   * [assertible.com](https://assertible.com) — Automated API testing and monitoring. Free plans for teams and individuals.
   * [bleemeo.com](https://bleemeo.com) - Free for 3 servers, 5 uptime monitors, unlimited users, unlimited dashboards, unlimited alerting rules.
+  * [Core Web Vitals History](https://punits.dev/core-web-vitals-historical/) - Find Core Web Vitals history for a url or a website.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -848,6 +848,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
   * [appspector.com](https://appspector.com/) - Mission control for remote iOS/Android/Flutter debugging. Free for small traffic usage (64MB of logs).
   * [assertible.com](https://assertible.com) — Automated API testing and monitoring. Free plans for teams and individuals.
   * [bleemeo.com](https://bleemeo.com) - Free for 3 servers, 5 uptime monitors, unlimited users, unlimited dashboards, unlimited alerting rules.
+  * [Core Web Vitals History](https://punits.dev/core-web-vitals-historical/) - Find Core Web Vitals history for a url or a website.
```
ripienaar, line 851: What benefit does this bring over the already really good services google and chrome provide?

What we're looking for here is SERVICES, like real value-add services, not something that runs on a personal blog url and where what the service does is a viable thing on its own right.

geeky-biz: > What benefit does this bring over the already really good services google and chrome provide?

Google's urls or chrome devtools do not show if a page's core web vitals improved / degraded in the last six months (only shows current performance). This tool allows you to enter your url to see if the speed degraded / improved in last six months. It pulls the data from Google's CrUX historical db.

> What we're looking for here is SERVICES, like real value-add services, not something that runs on a personal blog url and where what the service does is a viable thing on its own right.

- If you mean the service has to be owned by a company / corporate, then I agree - this is not suitable to be listed here.
- This service is used by dozens of developers every day to identify regressions in their core web vitals.
- What is service does is viable or no - it's based on Google's CrUX data (as listed on the linked page). Any one who can query Google's CrUX BigQuery can cross-check.

PR #3654, opened 2025-03-04
Remove IBM Cloud 25 GB Storage listing

data-tangles:
The Lite plan which offered 25GB Object Storage was deprecated as per https://www.ibm.com/docs/en/watsonx/saas?topic=storage-cloud-object-plans


ripienaar:
Thanks

PR #3655, opened 2025-03-06
Add Easy Site Search, remove Httpic - my previous startup

lgmys:
Easy Site Search is a much simpler alternative to Algolia, with drop in search widget and automated crawler. It allows unlimited search requests in the free plan, but only for up to 50 subpages on a website. 

Also, I am delisting my previous project Httpic.com as it is discontinued now, the domain points to some vpn service now.


ripienaar:
Thanks

PR #3656, opened 2025-03-06
Add Corsfix

reynaldichernando:
## Requirements


README.md
```diff
@@ -236,6 +236,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
   * [CometML](https://www.comet.com/site/) - The MLOps platform for experiment tracking, model production management, model registry, and complete data lineage, covering your workflow from training to production. Free for individuals and academics.
   * [Commerce Layer](https://commercelayer.io) - Composable commerce API that can build, place, and manage orders from any front end. The developer plan allows 100 orders per month and up to 1,000 SKUs for free.
   * [Conversion Tools](https://conversiontools.io/) - Online File Converter for documents, images, video, audio, and eBooks. REST API is available. Libraries for Node.js, PHP, Python. Support files up to 50 GB (for paid plans). The free tier is limited by file size and number of conversions per day.
+  * [Corsfix](https://corsfix.com) - CORS Proxy by Corsfix lets you fetch any resource on the web and bypass CORS errors. Free unlimited monthly requests for development.
```
ripienaar, line 239: Dpnt know about this, just seems like a Bad Idea. CORS exist for a reason.

reynaldichernando: hi @ripienaar , thank you for reviewing my PR

CORS proxy main goal is for fetching external (public and unauthenticated) resources that don't have CORS headers.

The use case is usually for people working with static websites (frontend only) where they don't have a server to perform the request from. If they had a backend, then they will perform their request via backend, as CORS is enforced in browser only.

It is not intended for solving misconfigured CORS on a server that they own. It is instead for fetching external resource they don't control.

some references:
- a popular open source cors proxy: https://github.com/Rob--W/cors-anywhere
- stackoverflow answer on cors proxy: https://stackoverflow.com/a/43881141/13781128
- article i wrote on cors proxy: https://corsfix.com/blog/cors-proxy

reynaldichernando: i know that this is a somewhat niche topic, but i'm happy to address any concerns or questions you might have

ripienaar: I asked a few Davs more up to speed with frontend work than me and everyone said the same. Cors exist for a reason and this is a bad idea. 

I am sure there are other lists this is better suited for, for me I will not accept it here

reynaldichernando: I understand, still, thanks for taking the time for reviewing and checking out CORS proxy 🙏

PR #3677, opened 2025-03-19
fynix

Kaushal3603:
## Requirements


README.md
```diff
@@ -535,6 +535,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
   * [Supermaven](https://www.supermaven.com/) — Supermaven is a fast AI code completion plugin for VSCode, JetBrains, and Neovim. Free tier provides unlimited inline completions.
   * [tabnine.com](https://www.tabnine.com/) — Tabnine helps developers create better software faster by providing insights learned from all the code in the world. Plugin available.
   * [v0.dev](https://v0.dev/) — v0 uses AI models to generate code based on simple text prompts. It generates copy-and-paste friendly React code based on shadcn/ui and Tailwind CSS that people can use in their projects. Each generation takes at minimum 30 credits. You start up with 1200 credits, and get 200 free credits every month.
+  * [Fynix](https://www.fynix.ai/) - Fynix combines real-time AI coding assistance with agent-powered code generation.
```
ripienaar, line 538: We only add entries once, please put it at the most appropriate place.
Please state what's free

Also worth noting I cant confirm what's free because the web page is broken on safari, says its an invalid address if I click getting started for free, so as is, this just won't make it on the list.

PR #3683, opened 2025-03-26
add TempMailSafe.com

hellodake:
add TempMailSafe.com

 

- [x]  It has a free tier not just a free trial
- [x]  Pricing information is clearly visible without signup or phone calls
- [x]  The submission mentions what is free
- [x]  The submission is not already present in the list
- [x]  The service has contact details of those running it and a privacy policy


ripienaar:
As per the text you were shown when opening this PR, we do not accept this kind of service anymore.

```
### Services we do not accept

   * Free dns services that are generic frontends to CloudFlare or similar
   * Services that are verbatim copy pastes of others while adding no value
```

PR #3702, opened 2025-04-17
Add time decimal converter

pandaupup:
## Requirements


ripienaar:
I am not sure this is really worth adding here, it has no API, no real purpose for developers that cant be solved already 100s of times over

pandaupup:
ok, thanks for your feedback

PR #3725, opened 2025-05-16
Remove SendGrid.com

juandbc:
SendGrid no longer offers a free tier; instead, it's now offering a 60-day free trial.
image


ripienaar:
Thanks

PR #3731, opened 2025-05-22
Update README.md

Harshit-shrivastav:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1295,6 +1295,7 @@ Update Time, five active automations, webhooks.
   * [Patr Cloud](https://patr.cloud/) — An easy-to-use cloud platform, among its paid services it offers to host three static sites for free.
   * [Serv00.com](https://serv00.com/) — 3 GB of free web hosting with daily backups (7 days). Support: Crontab jobs, SSH access, repositories (GIT, SVN, and Mercurial), support: MySQL, PostgreSQL, MongoDB, PHP, Node.js, Python, Ruby, Java, Perl, TCL/TK, Lua, Erlang, Rust, Pascal, C, C++, D, R, and many more.
   - [Sevalla](https://sevalla.com/) - Hosting platform designed to simplify the deployment and management of applications, databases, and static websites - 1GB site limit, 100GB free bandwidth, 600 free build minutes, 100 sites per account.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1295,6 +1295,7 @@ Update Time, five active automations, webhooks.
   * [Patr Cloud](https://patr.cloud/) — An easy-to-use cloud platform, among its paid services it offers to host three static sites for free.
   * [Serv00.com](https://serv00.com/) — 3 GB of free web hosting with daily backups (7 days). Support: Crontab jobs, SSH access, repositories (GIT, SVN, and Mercurial), support: MySQL, PostgreSQL, MongoDB, PHP, Node.js, Python, Ruby, Java, Perl, TCL/TK, Lua, Erlang, Rust, Pascal, C, C++, D, R, and many more.
   - [Sevalla](https://sevalla.com/) - Hosting platform designed to simplify the deployment and management of applications, databases, and static websites - 1GB site limit, 100GB free bandwidth, 600 free build minutes, 100 sites per account.
+  - [Webhostmost](https://www.webhostmost.com/free-web-hosting) — 125MB free storage, 1 email account, monthly login, 1 parked domain**, 1 database. Python, Redis, PHP, Node.js, WordPress.
```
ripienaar, line 1298: as per the PR template we dont accept these anymore:

```
```

Harshit-shrivastav: No, come on, it's not that. It's neither cPanel nor only PHP. You can review their website; they offer Node.js, Python, etc., and they are using a custom web panel instead of cPanel. They also guarantee 99.9 percent uptime, which cPanel hosting doesn't provide.

ripienaar: So it's custom and not DirectAdmin-based Web Control Panel?

Harshit-shrivastav: obviously, it's totally different from those cheap hosting services, i have already reviewed all the PR policy before posting anything

Harshit-shrivastav: Sorry, I am not getting you. For clarification, I can say that this website is not similar to Infinity Free or AeonHost, etc., which provide free PHP hosting with cPanel. It's totally different from them. Also, their web panel may be a fork or modification of DirectAdmin, but they are not using DirectAdmin or cPanel.

ripienaar: ```DirectAdmin-based Web Control Panel``` is what their site says.

Anyway, there are a million of these sites, we dont need to list them all.

Harshit-shrivastav: it's upto you if you want to list them but i would say there are only few websites that offer python and redis hosting

PR #3732, opened 2025-05-22
Change idx.dev with Firebase.studio

fgbbd:
Change name of IDX project to Firebase Studio, since IDX evolved recently to firebase studio.


ripienaar:
Thanks

PR #3733, opened 2025-05-23
Update Checkly's free monitoring offering

robinbowes:
I tried checkly, and found the number of free checks on the "Hobby" plan to be
quite limited, so went ot look for alternatives here and noticed their entry
was out-of-date.


ripienaar:
Thanks @robinbowes !

PR #3743, opened 2025-05-31
add LocalCert

wdhdev:
## Requirements

Hi there, I'm not 100% sure if this will be approved as it could be considered a generic frontend, however it is only a DNS service, with no public websites. It is similar to sslip.io which is already present in the list. The project allows users to host their internal applications on HTTPS subdomains.

I understand if this is not accepted.


ripienaar:
Useful but a bit limited, once you have UI you can add it again thanks

wdhdev:
@ripienaar Dashboard has been added: https://localcert.net/dashboard

More advanced features will be added soon as well.

PR #3750, opened 2025-06-05
Update CockroachLabs

anmoti:
The information was outdated and has been updated.
The official name of this service is CockroachDB Cloud.


ripienaar:
Thanks

PR #3751, opened 2025-06-06
Add fibo.dev for developers

abcdan:
A lot of these tools are paid or have issues with being not available as much as you want them to be (all the time). Fibo is free and hosted so it can auto scale during peak loads.

## Requirements


ripienaar:
Thanks

PR #3761, opened 2025-06-18
add WebMySite

Harshit-shrivastav:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -2045,6 +2045,6 @@ HTML Themes and UI Kits to create your applications faster!
   * [Killer Coda](https://killercoda.com/)  - Interactive playground in your browser to study Linux, Kubernetes, Containers, Programming, DevOps, Networking
   * [Axonomy App](https://axonomy-app.com/) - A free tool to create, manage and share your invoices with your clients. 10 free invoices per month.
   * [Table Format Converter](https://www.tableformatconverter.com) - A free tool to convert table data to different formats, such as CSV, HTML, JSON, Markdown and more.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2045,6 +2045,6 @@ HTML Themes and UI Kits to create your applications faster!
   * [Killer Coda](https://killercoda.com/)  - Interactive playground in your browser to study Linux, Kubernetes, Containers, Programming, DevOps, Networking
   * [Axonomy App](https://axonomy-app.com/) - A free tool to create, manage and share your invoices with your clients. 10 free invoices per month.
   * [Table Format Converter](https://www.tableformatconverter.com) - A free tool to convert table data to different formats, such as CSV, HTML, JSON, Markdown and more.
-
+* [WebMySite](https://webmy.site)– A no-code website builder offering forever-free templates and feature-rich modules like e-commerce, hotel booking, and LMS. Perfect for developers, tech agencies, and small businesses to create professional portfolios or full websites—no coding required.
```
ripienaar, line 2048: what's the pricing strategy here, currently its all free, what's the model? Free tier seems fully baked, typically that means shortly after the free tier gets massively cut in terms of capabilities etc.

So seems to me this is just a service thats very new and probably need to be a bit further along to be added here.

Remember the point of this list is services with free tiers for developers so that they can build solutions for customers who can then later rely on these services for a premium service.

Harshit-shrivastav: Okay, I noted the last paragraph.

PR #3777, opened 2025-06-30
include is-a.dev - a free subdomain provider

thepoetdj:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1301,6 +1301,7 @@ Update Time, five active automations, webhooks.
 
 ## Domain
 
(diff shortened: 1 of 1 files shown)


ripienaar:
As per the notice you had when submitting the PR we do not accept this kind of service anymore. 

Thanks.

thepoetdj:
@ripienaar shall we remove the **Domain** section in that case or do we want to keep them for historical purposes?

ripienaar:
No, the ones that are there are better quality

PR #3782, opened 2025-07-09
add LocalCert

wdhdev:
## Requirements

We have since added a [dashboard](
https://localcert.net/dashboard) and made various other improvements since our last PR: #3743

LocalCert is a 100% anonymous service and we do not collect any PII.


ripienaar:
Thank you

PR #3789, opened 2025-07-14
Add ExpenseEasy – privacy-first budgeting app

tariq235:
## Requirements


ripienaar:
Thanks, but I am afraid I do not think this fits the purpose of the list,

tariq235:
@ripienaar It's a direct competitor of Expensify. I am bit confused why Expensify is on the list then when both are similar product ?

ripienaar:
What's the SaaS offering your phone-installed APP offers? What's the (eg) API integration you provide? 

As this is a list for SaaS that's the question, maybe I missed it when reviewing your site so please share links

PR #3790, opened 2025-07-14
add Tmpmails

guaguaishushu:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -950,6 +950,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 ## Email
 
+  * [TmpMails](https://tmpmails.com/) - No registration required, get a temporary email address instantly. Protect your privacy and stay away from spam.
   * [10minutemail](https://10minutemail.com) - Free, temporary email for testing.
   * [AhaSend](https://ahasend.com) - Transactional email service, free for 1000 emails per month, with unlimited domains, team members, webhooks and message routes in the free plan.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -950,6 +950,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 ## Email
 
+  * [TmpMails](https://tmpmails.com/) - No registration required, get a temporary email address instantly. Protect your privacy and stay away from spam.
```
ripienaar, line 953: As per the PR template, we do not accept:

```
```

thanks

arnotongdev:
Thank you for your response. However, I believe that even if there are already sufficient options available, there should still be some new products added to maintain continuous updates of the product lineup. Moreover, our temporary email service is genuinely free without any quantity limitations, which still creates significant differentiation from other products. We hope to be included in your listing.

README.md
arnotongdev: @ripienaar Thank you for your response. However, I believe that even if there are already sufficient options available, there should still be some new products added to maintain continuous updates of the product lineup. Moreover, our temporary email service is genuinely free without any quantity limitations, which still creates significant differentiation from other products. We hope to be included in your listing.

ripienaar:
There are many other “awesome lists” I am sure you will find several where your services are a good fit.

arnotongdev:
Okay. Thanks for your reply.

PR #3798, opened 2025-07-20
add Greptile

kero13ro:
## Requirements


ripienaar:
This requires a form to be filled in for OSS repos which presumably have to be approved based on some unknown criteria.  Other services that support OSS take a GitHub onboarding process that check if the repo is open source.

So we've generally been denying free services that require this kind of form based signup, so we'ill reject this one also

PR #3811, opened 2025-08-06
Add Syncmygroup -> Sync Users and Groups across tools

Falit:
## Requirements


ripienaar:
Website seems pretty broken on Safari, probably a good signal to skip this one

Falit:
@ripienaar  thanks for bringing that up. It's been fixed.

ripienaar:
menu still  broken on vanilla safari  with all extensions off, think I will pass on this thanks.

PR #3822, opened 2025-08-21
Remove Fern

itsjuoum:
Removes Fern (buildwithfern.com) as they seem to have discontinued their free plan


itsjuoum:
Never mind, seems like it just isn't covered in the pricing page nor the docs as per [this issue](https://github.com/fern-api/fern/issues/5925).

ripienaar:
they must had free tier docs on pricing page and so must have removed it, in that case they dont fit this list anymore.

We require clear pricing information, so we can remove it.

PR #3824, opened 2025-08-22
remove shipyardapp.com

sayahosea:
shipyardapp.com redirects to alli.pmg.com, a digital marketing intelligence platform, so i assume the low-code service has stopped...?
i'm not really sure since their docs is still up (https://www.shipyardapp.com/docs).


ripienaar:
Thank you

PR #3826, opened 2025-08-26
Add Voice Clone to Generative AI section

tanchaowen84:
## Requirements

---

## Submission

Add [Voice Clone](https://voice-clone.org/) to the Generative AI section.

Voice Clone is an AI voice-cloning SaaS that enables instant generation of natural-sounding speech from your own voice.  
**Free plan:** 1,000 characters per day


ripienaar:
Does 1000 characters mean like “abc” is 3 characters?

if so that’s not useful

tanchaowen84:
Unfortunately, due to cost constraints, that’s indeed the case. However, this product also includes other tools—one of them is an Audio Enhancer, which I specifically built to improve audio quality and remove background noise. I noticed that most audio enhancement tools on the market require payment, which can be inconvenient. My tool is completely free to use, with no login required.

Perhaps this tool is a better fit for your list. You can find it here: [https://voice-clone.org/tools/audio-enhancer]. If you see value in it, I’d love the chance to resubmit.

ripienaar:
Yes most AI tools doesn’t really fit this list

you can add a new entry for the other tool.

tanchaowen84:
thks for replying, i would add a new pr for another tool then

ripienaar:
Thanks

PR #3836, opened 2025-09-09
Add ArchFormation to Cloud management solutions

Neifn:
## Requirements


ripienaar:
Pricing information and methodology must be clear ON THE SITE, this list is maintained by countless people who visit the sites and see if the information is valid and correct.  They do not come back to this PR to see explanations for how pricing works.

If the site does not explain its pricing policy its not going to be listed here.

Neifn:
Thank you for clarification, I will update documentation to include detailed explanation of tokens, and make a new request.

PR #3839, opened 2025-09-11
Remove CodeBeat from the Code Quality section

Shuudy:
Removed (https://codebeat.co) as the service has apparently shut down. 
Indeed, the homepage now shows a message that the app is discontinued.


ripienaar:
Thank you

PR #3846, opened 2025-09-18
Update README.md

luca-petre:
updated StateMesh entry in order to match with the other entries

## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,19 +171,6 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
     * [Queues](https://developers.cloudflare.com/queues/) - 1 million operations per month
     * [TURN](https://developers.cloudflare.com/calls/turn/) – 1TB of free (outgoing) traffic per month.
    
-  * [StateMesh](https://statemesh.net/)
-    * [Applications](https://docs.statemesh.net/applications/intro) - Free tier to deploy containerized web apps and APIs with 1 CPU, 512MB RAM, and autoscaling support.
-    * [Storage Providers](https://docs.statemesh.net/applications/storage-providers) - Attach persistent encrypted & replicated storage; options include built-in StateMesh volumes, block storage on the network, or your own S3-compatible bucket
-    * [Blockchain](https://docs.statemesh.net/blockchain/info) - EVM-compatible Layer 1 chain utilizing Fantom's Lachesis DAG consensus for high throughput and fast finality; supports MESH staking, cross-chain stablecoin bridging, and decentralized node state validation via Time Towers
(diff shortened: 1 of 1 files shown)


ripienaar:
We only list providers once, except the major cloud providers at the top of the list - I doubt state mesh would qualify

please read this and read the intent here. Every company gets a short single line entry. Please do like everyone else

luca-petre:
can you tell me where to place it? i would be more than glad to make the changes again. is it under the cloud management solutions section?

ripienaar:
Create a single single line entry that isn’t too long. State what’s free. 

There is over a thousand examples here on tbe list, it cannot be more clear.

README.md
```diff
@@ -194,6 +181,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
...
   * [deployment.io](https://deployment.io) - Deployment.io helps developers automate deployments on AWS. On our free tier, a developer (single user) can deploy unlimited static sites, web services, and environments. We provide 20 job executions free per month with previews and auto-deploys included in the free tier.
+  * [StateMesh](https://statemesh.net/) - StateMesh helps developers deploy containerized web apps and APIs on a decentralized cloud. On the free tier, a developer can run apps with 1 CPU and 512MB RAM, use managed PostgreSQL databases, attach persistent storage, launch virtual machines, and [deploy from GitHub with one click](https://console.cloud.statemesh.net/deploy?appname=sample&repository=https://github.com/state-mesh/samples.git). The free tier also includes CLI management, logging, monitoring, and access to open-source marketplace templates.
```
ripienaar, line 184: I do not see any information about a free tier on the pricing page, can you please share a link?

luca-petre: StateMesh does not explicitly advertise a free tier. However, the platform offers several features at no cost, which can be considered part of its free offerings. StateMesh provides certain features that are free to use indefinitely. These include high-performance ingress, custom resource management, instant scaling, and the ability to bring your own storage. Additionally, external ingress rules, such as HTTP/S and TCP support, dynamic routing, public endpoints, and custom domain names, are also free forever. StateMesh employs an active billing model, meaning you are only charged when your code is actively running. You won't incur charges while your code is waiting on external resources like API calls, database queries, or AI model responses. In the contrary, you also get paid if you become a node operator as well as referring friends, meaning you can gain back what you are charged for. Also, if the app is small enough, it can be free to maintain.

ripienaar: Pricing information has to be clear on your site.

This list is maintained by many people, they have to be able to confirm the claims of a free tier.

I think you're just not suitable to this list.

luca-petre: Okay, no worries. Thank you for your time!

PR #3854, opened 2025-09-26
Add Juno

peterpeterparker:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1208,6 +1208,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
   * [Flutter Flow](https://flutterflow.io) — Build your Flutter App UI without writing a single line of code. Also has a Firebase integration. The free plan includes full access to UI Builder and Free templates.
   * [getstream.io](https://getstream.io/) — Build scalable In-App Chat, Messaging, Video and audio, and Feeds in a few hours instead of weeks
   * [hasura.io](https://hasura.io/) — Hasura extends your existing databases wherever it is hosted and provides an instant GraphQL API that can be securely accessed for web, mobile, and data integration workloads. Free for 1GB/month of data pass-through.
+  * [Juno](https://juno.build) — Open-source, next-gen serverless platform to build, deploy, and run modern apps with the privacy and control of self-hosting. Free plan includes one managing and one project container.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1208,6 +1208,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
   * [Flutter Flow](https://flutterflow.io) — Build your Flutter App UI without writing a single line of code. Also has a Firebase integration. The free plan includes full access to UI Builder and Free templates.
   * [getstream.io](https://getstream.io/) — Build scalable In-App Chat, Messaging, Video and audio, and Feeds in a few hours instead of weeks
   * [hasura.io](https://hasura.io/) — Hasura extends your existing databases wherever it is hosted and provides an instant GraphQL API that can be securely accessed for web, mobile, and data integration workloads. Free for 1GB/month of data pass-through.
+  * [Juno](https://juno.build) — Open-source, next-gen serverless platform to build, deploy, and run modern apps with the privacy and control of self-hosting. Free plan includes one managing and one project container.
```
ripienaar, line 1211: while the pricing system is as clear as mud, it seems this is not a free tier.  You get a one off credit to build one thing and that's it.

To qualify - and be useful for development - you would need to be able to create, recreate etc, experiment and become familiar with the system, something I do not think a single one off credit would do.

Anyway, regardless of how useful or not, a one off credit does not quality for the list

peterpeterparker: Thanks a lot for the feedback @ripienaar 

Just to clarify: with the free tier you can create and experiment as you are mentioning, i.e. containers are spun up with resources that let you get familiar with and, for example, even run a website for a while.

It’s true that if you destroy your container you can’t create one additional for free but, you can reset the container you got and you do get free resources to start building.

That said, it’s of course up to you, and thanks again for considering it!

ripienaar: Sounds too restrictive to me, the spirit of free tiers is its a ongoing tier of free usage enough to let you explore the core feature set.

For example I might want to automate creation of project, containers etc, and free tiers let me do that because they renew fully monthly - in yours it's a one try and the free is over. 

So its more like a one time applied free credit, which isnt acceptable here

peterpeterparker: Fair enough. Thanks again for considering it and feedback.

PR #3855, opened 2025-09-26
Add Capawesome

robingenz:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 ## Source Code Repos
 
   * [Bitbucket](https://bitbucket.org/) — Unlimited public and private Git repos for up to 5 users with Pipelines for CI/CD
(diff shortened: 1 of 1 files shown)


ripienaar:
This is kind of very single technology orientated, we tend to not accept these - for example we've in the past rejected cloud hosting that only worked with Python etc.  So this is the same.

Further the entry is just incomplete and you clicked boxes that you made no effort to actually do what they say,.

robingenz:
>  The submission mentions what is free

You probably mean that. I wasn't sure if that meant the pricing page, as there was no section for this information in the issue template. Sorry, I have updated my branch.

I just saw that [ionicframework.com](https://ionicframework.com/appflow) is on the list which offers the same service as us.

Nevertheless, thank you for taking a look at my PR.

ripienaar:
Today I would say they would not qualify for the list - granted frontend isnt my space but they do seem to state they cater for multiple technologies, where your site is framed as being one technology.  Perhaps it's for more, but it's not clear.

ionic doesn't have a clear pricing page so should be removed

PR #3858, opened 2025-09-28
Add CarAPI.dev - Automotive Data API

melnikovic:
- Added CarAPI.dev to APIs, Data and ML section
- Comprehensive automotive data API with 9 endpoints
- Free tier: 100 requests/month permanently
- Includes VIN decode, stolen vehicle check, valuation, and more

## Requirements

 * [ X ] This is Software as a Service not self hosted
 * [ X ] It has a free tier not just a free trial
 * [ X ] Pricing information is clearly visible without signup or phone calls
 * [ X ] The submission mentions what is free
 * [ X ] The submission is not already present in the list
 * [ X ] The service has contact details of those running it and a privacy policy


ripienaar:
Thanks

PR #3867, opened 2025-10-06
remove non-functional sites

KHROTU:
feel free to double check anything, i live in sg so some site might not work for the region but not very likely:
- removed https://www.momentaryemail.com/ (site no longer up)
- removed https://designer.mocky.io/ (404, ssl self-signed)
- removed https://canister.io/ (site no longer up)
- removed https://touchlead.app/ (site no longer up)
- removed https://sourcelevel.io/ (site no longer up)
- removed https://libreqr.com/ (403)
- removed https://wishkit.io/ (site no longer up)
- removed https://mergify.io/ (1016, site no longer up)
- removed https://teletype.oorja.io/ (site no longer up)
- removed https://www.lovelylanding.net/ (site no longer up)
- removed https://www.mobile-patterns.com/ (site no longer up)
- removed https://www.pageshare.dev/ (site no longer up)
- removed https://easysitesearch.com/ (site no longer up)
- removed https://www.coingecko.com/en/api (site no longer up)
- removed https://www.lecturify.net/index.en.html (timeout, site no longer up)
- removed https://osdn.net/ (site no longer up)
- removed https://terminusdb.com/pricing (522, site no longer up)
- removed https://www.pivotaltracker.com/ (site no longer up)
- removed https://alerty.ai/ (site no longer up)
- removed https://jsonserve.com/ (site no longer up, ssl self-signed, broken redirect)
- removed https://usequalli.com/ (site no longer up)
- changed https://redislabs.com/try-free/ to https://redis.io/
- removed https://openapi.security/ (404, ssl self-signed)
- removed https://layercode.com/usecsv (changed to ai thing)
- removed https://glassfy.io/ (site no longer up)
- changed https://rive.app/community/ to https://community.rive.app/c/start-here/
- removed https://formpost.app/ (site no longer up)
- removed https://markdowntools.com/ (timeout)
- removed https://mirna.cloud/ (site no longer up)
- removed https://hyperforms.app/ (hsts and certificate expired)
- removed https://kandi.openweaver.com/ (502)
- removed https://haveibeenpwned.com/ (site no longer up)
- removed https://drv.tw/ (site no longer up)
- removed https://dicloud.net/trackwith-privacy-focused-analytics/ and https://dicloud.net/sweetuptime-server-uptime-monitoring/ (the entire site, except the home page, is non-functional)
- removed https://cypress.preflight.com/ (site no longer up)
[...]

PR #3869, opened 2025-10-07
Revert "remove non-functional sites"

ripienaar:
Reverts ripienaar/free-for-dev#3867


FatGrizzly:
@ripienaar do you think something like https://github.com/lycheeverse/lychee-action would be helpful?

ripienaar:
We've had this sort of thing in the past, it was quite short lived due to false hits, I ended up having to just check everything :)

PR #3879, opened 2025-10-20
add MicrophoneTest website

arnotongdev:
## Requirements


ripienaar:
The site makes a big deal about it being securely encrypted etc - but also says its all local, seems a bit sensationalist :)

Looks useful but thats a bit missleading

arnotongdev:
The website really doesn’t save any data—all the processing happens right in your browser. We’re just testing your microphone directly in your browser. We take your privacy super seriously, but we also want to offer some handy tools. Thanks!

ripienaar:
That's correct so making a big deal about your encryption - when its all local - just seems pointless

arnotongdev:
That's the right way to see it. But not everyone might get it—some don't really understand how it works. So it's still important to stress why privacy matters, clear up their doubts, and make them feel more at ease using it.

madhuryadutta:
Hi @tommcflydev , I think @ripienaar  made a good point.
Since there is no server-side processing, there is no need to use words like “encrypted.”

If you want to tell users that the tool is privacy focused, then the tagline should be simple , like -

- 100% local. Your audio never leaves your device.
- Runs entirely in your browser. No uploads. No data stored.

PR #4288, opened 2026-04-11
removing an offer that is no longer any good

brandonzylstra:
Serv00 is not taking new customers.  Whether this was a legitimate offer at the time it was added, I don't know.  But right now it amounts to bait and switch as long as it is left in this list.

image


ripienaar:
Thank you

brandonzylstra:
You're very welcome!  Thanks for maintaining the list!

PR #4293, opened 2026-04-11
Add FastDevTool — 50+ free browser-based developer utilities

o0flame:
FastDevTool (https://fastdevtool.com) is a collection of 50+ free developer utilities that run entirely in the browser — JSON formatter/beautifier/validator, Base64, URL encoder/decoder, JWT decoder, Regex tester, UUID generator, Hash generator, SQL formatter, YAML/XML tools, and more.

**Why it belongs here:**
- Completely free, no account required, no paid tier
- No server processing — all computation runs client-side
- Uses Monaco editor (VS Code engine)
- Offline capable via PWA service worker
- Shareable URLs via pako compression (no server storage)

Added alphabetically after FormatJSONOnline.com in the F section.


ripienaar:

PR #4294, opened 2026-04-11
Add ApexToWWW to DNS section

x2q:
## What is ApexToWWW?

[ApexToWWW](https://www.apextowww.com/) is a free DNS apex/root/naked domain to www subdomain 301 redirect service.

## Why add it?

Many developers need to redirect their apex domain (example.com) to www (www.example.com) because hosting platforms like GitHub Pages, Netlify, and Vercel require CNAME records, which can't be set at the DNS root. ApexToWWW solves this with a free, no-signup service.

## Features (all free, no limits)

- 301 permanent redirect with full path/query preservation
- Automatic SSL via Let's Encrypt
- IPv6 (dual-stack) support
- HTTP/3 (QUIC) support
- No account or signup required
- <50ms redirect latency

## How it works

Users add two DNS records (A + AAAA) to their apex domain, and ApexToWWW handles the redirect automatically.


ripienaar:

PR #4295, opened 2026-04-11
Railway no longer has a free tier

brandonzylstra:
instead they have a free 30-day trial, and the formerly free tier now costs $5/month

> * [ ] Large Language Models and other AI tick this box

that's clever... I didn't notice that before 😅


ripienaar:
Thank you

PR #4351, opened 2026-05-04
Fix ToC link formatting

BunnyNabbit:
Fixes the "Back to top" link being inlined with the Metashot entry by adding necessary whitespace for it to be interpreted as a paragraph.

## Requirements


ripienaar:
Thank you

PR #4360, opened 2026-05-06
Add DevBox to Miscellaneous

aifly:
DevBox is a collection of 35+ free developer tools. All tools run client-side with no signup required.

## Requirements


ripienaar:
we do not accept further generic toolboxes.

PR #4376, opened 2026-05-10
Add Laravel CSP Generator

itxshakil:


ripienaar:
Bit too single tech specific for this list sorry. Thank you.

itxshakil:
> Bit too single tech specific for this list sorry. Thank you.

Sure, Although it generates header too that can be used by any language.

PR #4386, opened 2026-05-17
Remove Claw Cloud from list

hesprs:
Claw Cloud Run is discontinued: https://question.run.claw.cloud/questions/10010000000003261. And the rest of Claw Cloud is no longer free.

This is a deletion rather than an addition. So I don't need to include PR checkboxes.


ripienaar:
Thank you

PR #4398, opened 2026-05-24
Add Mello to Tools collection

MelloTools:
Mello is a directory of ~120 browser-only utility tools (JSON, image, PDF, password, color, etc.). Free, no signup, no upload.

## Requirements

 * [ X ] This is Software as a Service not self hosted
 * [ X ] It has a free tier not just a free trial
 * [ X ] Pricing information is clearly visible without signup or phone calls
 * [ X ] The submission mentions what is free
 * [ X ] The submission is not already present in the list
 * [ X ] The service has contact details of those running it and a privacy policy


MelloTools:
Just noticed I forgot to check the boxes, mind re-checking? @ripienaar

ripienaar:
We don’t accept browser utilities. So neither would we accept a list of them.

MelloTools:
Got it, thanks for the quick reply!

PR #4413, opened 2026-05-29
Remove Leapcell (end of life)

Awesome-E:
Remove Leapcell, as they are sunsetting hosting.

Unfortunately, there is no public page documenting this (this change was fairly obscured), but it should be removed so people don't create an account thinking the service is still active.

image

## Requirements


ripienaar:
Thanks

PR #4444, opened 2026-06-08
Update Forminit free trier from 50 to 100 submissions per month

wssn:
## Update

This PR updates the existing Forminit listing to reflect the current free tier.

The free submission limit has increased from 50 to 100 submissions per month.

Original listing PR: #4349


ripienaar:
Thank you

PR #4476, opened 2026-06-13
Update Oracle Cloud A1 Compute to reflect new Always Free limits

passajked:
Oracle has recently updated their Always Free limits to reduce the A1 cores from 4 to 2 and A1 memory from 24GB to 12GB.

From https://docs.oracle.com/en-us/iaas/Content/FreeTier/freetier_topic-Always_Free_Resources.htm

>OCI Ampere A1 Compute instances (Arm processor): All tenancies get the first 1,500 OCPU hours and 9,000 GB hours per month for free for VM instances using the VM.Standard.A1.Flex shape, which has an Arm processor. For Always Free tenancies, this is equivalent to 2 OCPUs and 12 GB of memory.

## Requirements


ripienaar:
Thank you

PR #4477, opened 2026-06-14
Add LeadFinder API

Perufitlife:
## Requirements

Adds LeadFinder to the APIs, Data, and ML section. It's a hosted REST API that returns local business leads (name, phone, website, address, rating, category) pulled from Google Maps by niche and city. Free tier is 10 requests/day with no API key, and the response is plain JSON so it's easy to test from a browser or curl. Pricing for the full export is shown on the site without signup.


ripienaar:
Thanks, doesn't really fit the scope of the list though

PR #4491, opened 2026-06-18
add AIVA to Testing section

vashek:
Disclosure: I work on AIVA.

## Requirements

 * [x] Pricing information is clearly visible without signup or phone calls: see https://app.aiva.works/about?page=pricing
 * [x] The service has contact details of those running it and a privacy policy: see https://app.aiva.works/about?page=contact

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,6 +546,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 ## Testing
 
+  * [AIVA](https://aiva.works/) - No-code visual testing for web apps - tests the page the way a human sees it, no DOM, selectors, or XPath. Run your first test without signing up. Free tier available; paid plans add parallel test-execution agents.
   * [Appetize](https://appetize.io) - Test your Android & iOS apps on this Cloud Based Android Phone / Tablets emulator and iPhone/iPad simulators directly in your browser. The free tier includes two concurrent session with 30 minutes of usage per month. No limit on app size.
   * [Argos](https://argos-ci.com) - Open Source visual testing for developers. Unlimitedprojects, with 5,000 screenshots per month. Free for open-source projects.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -546,6 +546,7 @@ This list results from Pull Requests, reviews, ideas, and work done by 1600+ peo
 
 ## Testing
 
+  * [AIVA](https://aiva.works/) - No-code visual testing for web apps - tests the page the way a human sees it, no DOM, selectors, or XPath. Run your first test without signing up. Free tier available; paid plans add parallel test-execution agents.
```
ripienaar, line 549:
```
Yes! AIVA is built on a freemium model currently in free beta. Every user has access to a powerful free tier that's genuinely useful. You can start testing immediately and explore all the core features. In the future, we'll introduce premium and enterprise features, but we promise transparency every step of the way.
```

There's no long term commitment to a free tier here, so once you're out of beta and have transparent, real, pricing feel free to open a new PR

vashek: In case you're willing to reconsider, the pricing page does make it clear what we intend to keep free even after we launch billing.

ripienaar: Probably missing it but do you mind pasting here where it says that?

I only see presently it’s free while in beta and freemium model and pricing is what you consider might one day be the pricing but not clearly stating intention to have a free tier long term

vashek: To be fair, I guess it doesn't say that explicitly - that's actually good feedback, thanks. :) (Now I sound like an AI. I promise I'm not. ;))

The [pricing page](https://app.aiva.works/about?page=pricing) does show a free tier called Free and then that the Team tier is "Free in early access", which we meant to imply that the tier actually called Free will stay free even after early access. I guess we could make it clearer.

Anyway, thanks for reconsidering. You'd make my day. If you decide against, no hard feelings, it's your list.

ripienaar: The background is that maintaining this list with so many entries is a big task and to help everyone who contribute we need theee things to be clear. 

So once your pricing is set feel free to open another PR

PR #4497, opened 2026-06-20
Remove defunct services (CheckAPI, SOOS, MetricsWave, Userbird)

loginov-kirill:
Removed CheckAPI, SOOS, MetricsWave, and Userbird because they are dead or defunct:

- CheckAPI: DNS resolution fails
- SOOS: HTTP 410 Gone
- MetricsWave: DNS resolution fails
- Userbird: connection timeout / redirects to broken site


ripienaar:
Thank you

PR #4512, opened 2026-06-26
Add InfinityFree to the free domain list

ert11er:
## Requirements

InfinityFree is a great hosting site which gives a free subdomain as well. Though it has php+mysql free hostings, it has normal hostings and php+mysql is optional and can be closed. If it only had PHP+Mysql hosting, I wouldn’t send this PR


ripienaar:
We don't really accept any more sub domain services anymore - they have to be very high quality and offer full NS record delegation and high quality dashboards.

This site, with its 4 sides of ads that's almost bigger than the site itself does not qualify.

ripienaar:
Thank you

PR #4534, opened 2026-06-30
Added Chrome Remote Desktop

a12dongithub:
## Requirements

### DIFF
--- a/README.md
+++ b/README.md
@@ -1632,7 +1632,7 @@ Update Time, five active automations, webhooks.
   * [Getscreen.me](https://getscreen.me) -  Free for 2 devices, no limits on the number and duration of sessions
   * [RemSupp](https://remsupp.com) - On-demand support and permanent access to devices (2 sessions/day for free)
   * [RustDesk](https://rustdesk.com/) - Open source virtual/remote desktop infrastructure for everyone!
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1632,7 +1632,8 @@ Update Time, five active automations, webhooks.
   * [Getscreen.me](https://getscreen.me) -  Free for 2 devices, no limits on the number and duration of sessions
   * [RemSupp](https://remsupp.com) - On-demand support and permanent access to devices (2 sessions/day for free)
   * [RustDesk](https://rustdesk.com/) - Open source virtual/remote desktop infrastructure for everyone!
-
+  * [ChromeRemoteDesktop](https://remotedesktop.google.com/) - Free and practically no limit on number of devices, owned by Google so needs a Google account.
```
ripienaar: Please add it by the other Google things at the top 

We lost a company only once

a12dongithub: Thank you for your comment, and I can do that but I believe chrome remote desktop fits better in the remote desktop category as it is a standalone application released by google, and not part of the google cloud. If you still think I should change i will do that.

ripienaar: It might be better, I am still not adding the same company more than once. It's impossible to make a nuanced decision for each of thousands of submissions and someone will always feel they are on the wrong side of the line.

a12dongithub: Done!

ripienaar:
Thank you

PR #4560, opened 2026-07-05
Add System Status monitoring service

sudo-rsingh:
https://status.silinode.in/

## Requirements


ripienaar:
This looks a bit brand new - you'd expect a robust service hosted on something.example.com to have stuff on example.com introducing product etc

It's a good service to be clear but I think once you're more established and the product is done and have a future payment information up you should come and add it then

sudo-rsingh:
Thanks for the feedback.

PR #4587, opened 2026-07-15
Add Funding Signals to APIs, Data and ML

labombas:
Adding my API, Funding Signals. It tracks new SEC Form D filings every day, filters out amendments and investment funds, and gives back a scored list of companies that just raised money, with contact info. Free tier works without an API key.

## Requirements


ripienaar:
Thanks. Don’t think it fits the scope of the list really.

PR #4591, opened 2026-07-17
Add Ceki Browser Rental — real residential Chrome for AI agents

iWedmak:
Rent real Chrome browsers per minute (/bin/bash.01/min), real residential IPs, bypass captchas. Free signup, pay-as-you-go pricing.

## Requirements


ripienaar:
Thanks but I don’t think it’s a good fit for this list

PR #4606, opened 2026-07-19
Update Rivestack free tier description

geekyfox90:
I'm the founder of Rivestack. Our current entry says the free tier includes daily snapshots and 14 day point in time recovery, but the free tier does not include backups. This updates the description to what the free tier actually offers, so readers are not promised a restore feature that is not there. One line changed, no other entries touched.


ripienaar:
It was closed because against contributor guide, aup and PR template you used Claude to contribute here. 

Claude also ignored the CLAUDE.md - or more likely you told it to continue anyway because I test this and Claude refuse to contribute here on its own 

I will be removing your entry

geekyfox90:
Fair point. I used AI to make the edit because I was trying to avoid a human error, but I forgot to read the contribution rules first. Thats on me, sorry about that. I understand why you closed it and it wont happen again.

Thanks for maintaining free-for-dev.

PR #4618, opened 2026-07-20
Update the Google Gemini API entry

stefan9999991:
Update the Google Gemini API entry, which was renamed to Google AI Studio. Also updated the sadly worsened offer

 * [x] Large Language Models and other AI tick this box


ripienaar:
Thanks

PR #4625, opened 2026-07-21
Fix monitoring list formatting

BunnyNabbit:
Fixes indentation of two entries in the monitoring list, so elements aren't interpreted as child elements.

## Requirements


BunnyNabbit:
I'm noticing another issue in the same list. expanding scope of PR to include other fixes.

PR #4626, opened 2026-07-21
Adding FreeHosts

8B-Class:
## Requirements

FreeHosts is a community-curated directory that helps developers quickly find free hosting providers for websites, apps, bots, and databases — with community reviews and feature comparisons to save time versus researching each provider individually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1661,6 +1661,8 @@ Update Time, five active automations, webhooks.
 
 ## Other Free Resources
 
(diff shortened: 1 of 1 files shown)


ripienaar:
We don’t tend to reference other lists. We have done old entries which I would rather remove then add more. 

Thanks.

8B-Class:
> We don’t tend to reference other lists. We have done old entries which I would rather remove then add more.
> 
> Thanks.

May I know the reason behind it?

8B-Class:
From what I have been following the list for a while, the list seems to be up to date and helping multiple developers. So I thought to get added to help more developers 

I am not associated or involved in FreeHosts in any way.

PR #4658, opened 2026-07-27
Add Shadcn Dashboard and Shadcn Space in Design and UI

vaibhav-gupta5:
## Requirements


ripienaar:
These are not services. I know we have the odd other similar entry here but ultimately I don’t think they belong so not adding more like it. Thanks.

PR #4709, opened 2026-08-10
Add Rclone

asieftejani:
Unfortunately its self-hosted but very useful for developers (like vscode)
Wondering if you can make an exception (or maybe create a seperate self-hosting section)

## Requirements


ripienaar:
This list is for SaaS. Thanks.

PR #4750, opened 2026-08-20
Add Sample Files resource to README

mzeeshanid:
Added a new resource i.e. [Sample Files](https://mzeeshan.me/tools/sample-files), providing collection of use cases based testing files in different formats helpful for testing and QA. Most online resources only provide size and resolution based variants which are not enough. For example, consider [MP4](https://mzeeshan.me/tools/sample-files/extensions/mp4), generally file size based and video quality based (HD, FHD, 4K) variants are available but codec based, bitrate based and metadata based variants are not available in one place. Generating files for these cases take times and deviate from the business logic of the app, use ready made files. Each file variant is created with its designated software and command line utility. 

## Requirements


ripienaar:
Thanks

PR #4751, opened 2026-08-20
Update GhostChat entry — free plan now includes an AI chatbot

jacoborbach:
Updating my own existing entry. As of 20 Aug 2026 the GhostChat free plan includes the built-in AI chatbot (10 AI-written replies/month) — previously it was a paid add-on, so the current line understates what the free tier offers. No other changes.


ripienaar:
Thanks. 

10/month isn’t a useful amount. Now the entry is just embarrassing tbh.

PR #4760, opened 2026-08-22
Update GhostChat: free AI allowance bumped from 10 to 25 replies/mo

jacoborbach:
GhostChat's free plan AI allowance is now 25 replies/month, up from 10, so the entry is out of date.

You mentioned on #4751 that 10/month wasn't a useful amount — it's 25 now.

Also added "No credit card required" to the same line.

No other entries touched.


ripienaar:
thanks

PR #4762, opened 2026-08-22
Remove AndroidIDE from IDE and Code Editing section

Joyanta-Kumar:
This project is archived and no longer maintained


ripienaar:
Thanks

PR #4845, opened 2026-09-09
Add FormatPixl to Design and UIAdd FormatPixl to resources list in README

SuffiyanKhan:
### Requirements

Contributing here is very easy, but does require attention to details.
We do not accept LLM written submissions.

### Description
FormatPixl is a 100% free, client-side in-browser image optimization and conversion toolkit. No signups, completely private, and runs entirely in the browser using Canvas APIs.


ripienaar:
We don’t accept these kinds of tools anymore. Thanks.

